PR #3, opened 2022-04-14
Add highlights to README

ZacSweers:


seankim-android (approved):
Are these highlights available somewhere in our internal repos? This is looking great and reads very well. I hope more folks can see the lot of work that you & devXp (& BR? & somewhat the greater android org? I don't know) have done in the gradle plugin.

ZacSweers:
Tbh it's definitely not something we've always been good about publicizing as it's mostly irrelevant to the average developer 😅

PR #5, opened 2022-04-19
Use finalizeDsl for libraryextension callback in mod stats

ZacSweers:


slack-plugin/src/main/kotlin/slack/stats/ModuleStats.kt
```diff
@@ -176,35 +177,36 @@ public object ModuleStatsTasks {
...
-        project.afterEvaluate {
-          val extension = the<LibraryExtension>()
-          val targetVariant =
-            if (multiVariant) {
-              val defaultBuildType = extension.buildTypes.find { it.isDefault }?.name ?: "release"
-              val defaultFlavor = extension.productFlavors.find { it.isDefault }?.name ?: ""
-              "$defaultFlavor${defaultBuildType.safeCapitalize()}"
-            } else {
-              "release"
+        project.configure<LibraryAndroidComponentsExtension> {
```
seankim-android, line 180: Read this as a reference: https://developer.android.com/reference/tools/gradle-api/7.1/com/android/build/api/variant/LibraryAndroidComponentsExtension

seankim-android (approved):

PR #6, opened 2022-04-19
Document improved properties system

ZacSweers:


README.md
```diff
@@ -68,6 +68,12 @@ their impact. We support both Intel and Apple Silicon macs now and contain this
 `ThermalsParser.kt`. This also includes helpful charting APIs for visualizing the data (courtesy
 of our friends at Square).
 
+### Better Properties
+
+Gradle's built-in property support is limited and has surprising behavior, so we have our own 
+system on top of it that has consistent precedence, `local.properties` support, local project 
+properties, and configuration caching support. Check out `safeProperty()` in `PropertyUtil.kt`.
```
seankim-android, line 75: Minor nit - `local.properties` support vs. local project properties. 

We explicitly mention `local.properties`, which makes sense, but I think “local project properties” includes `local.properties`.

Perhaps we can expand on the latter more or delete the former?

README.md
```diff
@@ -68,6 +68,12 @@ their impact. We support both Intel and Apple Silicon macs now and contain this
 `ThermalsParser.kt`. This also includes helpful charting APIs for visualizing the data (courtesy
 of our friends at Square).
 
+### Better Properties
+
+Gradle's built-in property support is limited and has surprising behavior, so we have our own 
+system on top of it that has consistent precedence, `local.properties` support, local project
```
seankim-android, line 74: Not - just a suggestion with simplified sentence from grammar tool

```suggestion
Gradle's built-in property support is limited and has surprising behavior, so our system has 
consistent precedence, local.properties support, local project properties, ...
```

seankim-android (approved):

README.md
ZacSweers: nope they are different things. One comes from `local.properties` files and the other comes from project-local `gradle.properties` files

README.md
ZacSweers: I don't think this wording is clearer tbh. More declarative tone yes, but calling out that we have our own system is sorta the focal point in this one

README.md
seankim-android: Thanks that makes sense. We can elaborate the latter with your response here with "project-local `gradle.properties` files", but this was just a nit.

PR #7, opened 2022-04-19
Unify thermals watching API

ZacSweers:
This unifies our two thermals logging APIs behind a common `ThermalsWatcher` that can be simply started/stopped in `SlackTools`. This makes it fairly simple to work with and possibly easy to ~copy into other projects~ release as a separate project some day :).

Now all the core logic lives in `ThermalsWatcher.kt`. Would like to some day move the polling logic to coroutines too. 

Opportunistically cleaned up the README doc around properties too per @seankim-android's previous PR comment

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/gradle/SlackTools.kt
+++ b/slack-plugin/src/main/kotlin/slack/gradle/SlackTools.kt
@@ -98,11 +98,12 @@ public abstract class SlackTools @Inject constructor(providers: ProviderFactory)
   override fun close() {
     // Close thermals process and save off its current value
     thermalsAtClose = thermalsWatcher?.stop()
-    if (!parameters.offline.get()) {
-      thermalsAtClose?.let { thermalsReporter?.reportThermals(it) }
-    }
-    try {} catch (t: Throwable) {
-      logger.error("Failed to parse thermals", t)
+    try {
+      if (!parameters.offline.get()) {
+        thermalsAtClose?.let { thermalsReporter?.reportThermals(it) }
+      }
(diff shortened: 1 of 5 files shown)


slack-plugin/src/main/kotlin/slack/gradle/SlackTools.kt
```diff
@@ -98,91 +91,19 @@ public abstract class SlackTools @Inject constructor(providers: ProviderFactory)
...
-      if (!parameters.offline.get()) {
-        thermalsAtClose?.let { thermalsReporter?.reportThermals(it) }
-      }
-    } catch (t: Throwable) {
+    // Close thermals process and save off its current value
+    thermalsAtClose = thermalsWatcher?.stop()
+    if (!parameters.offline.get()) {
+      thermalsAtClose?.let { thermalsReporter?.reportThermals(it) }
+    }
+    try {} catch (t: Throwable) {
```
seankim-android, line 104: Try block here is empty. Did you mean to put the lines of code above to this try block? Otherwise, we should just remove it.

ZacSweers: Oops, yeah that's a leftover

ZacSweers: [652e92a](https://github.com/slackhq/slack-gradle-plugin/pull/7/commits/652e92a801c377751ffb61bcd493bb77f38d2e55)

seankim-android (approved):
Lgtm!

PR #14, opened 2022-04-30
Upgrade to dependency analysis plugin 1.0

ZacSweers:
This updates DAGP to its 1.0 APIs! This included major updates to DependencyRake along the way, namely:
- Removing our leftover shims for resources-only projects (these should Just Work now that we use non-transitive R classes I think?)
- Removing our custom facade handling as this is now covered by the plugin's "bundles" notion
- Apply common bundles and configurations in SGP to projects

Will take this for a spin in the main repo when I'm back from my trip


ZacSweers:
Note I'll do a separate PR to address suggestions it has for this own repo

slack-plugin/src/main/kotlin/slack/gradle/SlackRootPlugin.kt
```diff
@@ -155,23 +155,30 @@ internal class SlackRootPlugin : Plugin<Project> {
...
-            onAny {
-              severity("warn")
-              exclude(
-                // This is a resources-only project.
-                ":libraries:l10n-strings"
-              )
-            }
-          }
-        }
+        issues { all { onAny { ignoreKtx(true) } } }
```
autonomousapps, line 158: btw, I _think_ you could use dot notation here? Or if not, it would be easy to support if you'd prefer that syntax for brevity.

autonomousapps (commented):
Neat, thanks for sharing this! I'd love to hear any thoughts you might have on the ergonomics of the new API.

ZacSweers:
Ok took it through its paces and fixed some things. Looking mostly good now other than some known issues around test fixtures support

seankim-android (approved):
LGTM. Thanks Zac for giving an overview over Zoom!

PR #21, opened 2022-05-12
Add support for sqldelight 2 in mod stats

ZacSweers:


gradle/libs.versions.toml
```diff
@@ -17,7 +17,6 @@ moshix = "0.17.1"
 nullawayGradle = "1.3.0"
 okhttp = "4.9.3"
 spotless = "6.5.0"
-sqldelight = "1.5.1"
```
ZacSweers, line 20: this was unused

PR #23, opened 2022-05-13
Update to Gradle 7.5-rc-1

ZacSweers:
Covers a buncha things!
- Kotlin 1.6 (they still target 1.4, but the embedded runtime is now 1.6)
- Update OkHttp to 5.x since we can use Kotlin 1.6 now
- Use new `resolvedArtifacts` API to properly wire configurations into task inputs
- Make `BaseDependencyCheckTask` and its subtasks compatible with configuration caching + fully cacheable
- Use `ProviderFactory#exec` everywhere for process work in order to stay compliant with configuration cache and also allow more lazy provider wiring
- Clean up a bunch of eager provider get() calls that were leftover (opportunistic)


ZacSweers:
Currently blocked on https://github.com/gradle/gradle/issues/20787

slack-plugin/src/main/kotlin/slack/gradle/tasks/BaseDependencyCheckTask.kt
```diff
@@ -15,53 +15,47 @@
...
+
+  protected abstract fun handleDependencies(identifiersToVersions: Map<String, String>)
+
+  protected fun configureIdentifiersToVersions(configuration: Configuration) {
+    identifiersToVersions.putAll(
+      configuration.incoming
+        .artifactView {
+          attributes { attribute(AndroidArtifacts.ARTIFACT_TYPE, ArtifactType.AAR_OR_JAR.type) }
+          lenient(true)
+        }
```
autonomousapps, line 38: It looks like you might only care about external dependencies? (based on the filter at line 47) In that case, consider this:
```kotlin
// Only resolve external dependencies! Without this, all project dependencies will get _compiled_.
view.componentFilter { id -> id is ModuleComponentIdentifier }
```
That's a code snippet from an internal plugin I maintain for resolving dependencies.

slack-plugin/src/main/kotlin/slack/gradle/tasks/CheckDependencyVersionsTask.kt
```diff
@@ -37,18 +39,21 @@ public abstract class CheckDependencyVersionsTask : BaseDependencyCheckTask() {
 
   @get:Input public abstract val mappedIdentifiersToVersions: MapProperty<String, String>
 
+  // Only present for cacheability
```
autonomousapps, line 42: 👍 It can also be useful for auditing the task and for writing tests, since it's easier to assert against file contents vs console output.

slack-plugin/src/main/kotlin/slack/gradle/tasks/PrintFossaDependencies.kt
```diff
@@ -36,6 +42,7 @@ import org.gradle.api.tasks.OutputFile
  *
  * This task is not cacheable as it has no inputs and is run on demand.
```
autonomousapps, line 43: This kdoc looks out of date now.

gradle/libs.versions.toml
```diff
@@ -15,7 +15,7 @@ mavenPublish = "0.19.0"
 moshi = "1.13.0"
 moshix = "0.17.1"
 nullawayGradle = "1.3.0"
-okhttp = "4.9.3"
+okhttp = "5.0.0-alpha.7"
```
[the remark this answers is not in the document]
ZacSweers: yessir

gradlew
```diff
@@ -205,6 +205,12 @@ set -- \
         org.gradle.wrapper.GradleWrapperMain \
         "$@"
 
+# Stop when "xargs" is not available.
+if ! command -v xargs >/dev/null 2>&1
+then
+    die "xargs is not available"
+fi
+
```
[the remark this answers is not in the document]
ZacSweers: yup

slack-plugin/src/main/kotlin/slack/ProcessExtensions.kt
```diff
@@ -16,47 +16,49 @@
...
+      // Apparently Gradle wants us to distinguish between the executable and its arguments, so...
+      // we try to futz that here. But also this is silly.
+      commandLine(arguments[0])
+      args = arguments.drop(1)
+      inputWorkingDir?.let { workingDir(it) }
+    }
+    .standardOutput
+    .asText
+    .map { it.trimAtEnd() }
+    .map { it.ifBlank { sneakyNull() } }
```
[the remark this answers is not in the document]
ZacSweers: Kotlinc will error if you try to explicitly return null here even though gradle allows it because it has no reasonable annotations around this. This abuses kotlin contracts to promise the compiler it's non-null (and thus remove the runtime nullcheck + compile safely) while still actually using null at runtime anyway for gradle

slack-plugin/src/main/kotlin/slack/gradle/ApkVersioningPlugin.kt
```diff
@@ -57,29 +57,27 @@ internal class ApkVersioningPlugin : Plugin<Project> {
...
-      // TODO eventually make this computed again, eagerly resolving to avoid
-      //  https://github.com/gradle/gradle/issues/19658
+      val versionMajor = project.localGradleProperty("versionMajor")
+      val versionMinor = project.localGradleProperty("versionMinor")
+      val versionPatch = project.localGradleProperty("versionPatch")
+
       val user =
         project
           .ciBuildNumber
           .map { "" }
```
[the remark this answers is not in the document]
ZacSweers: that's... a good point. I'm not sure what's up with this, I'll look at it more in a followup

slack-plugin/src/main/kotlin/slack/gradle/tasks/BaseDependencyCheckTask.kt
ZacSweers: oh neat, thanks

slack-plugin/src/main/kotlin/slack/gradle/tasks/PrintFossaDependencies.kt
ZacSweers: [280d9ef](https://github.com/slackhq/slack-gradle-plugin/pull/23/commits/280d9ef3bb000b1fb66aff79e68b5ace22737c8f)

slack-plugin/src/main/kotlin/slack/gradle/tasks/BaseDependencyCheckTask.kt
ZacSweers: Does this look right? [ef0ad80](https://github.com/slackhq/slack-gradle-plugin/pull/23/commits/ef0ad804c62767660311d612de744c173d82cb90)

autonomousapps (approved):
LGTM!

PR #24, opened 2022-05-15
Kotlin 1.7.0 prep

ZacSweers:
- Use the new kotlin base plugin as a hook
- Add a toe-hold for supporting the K2 compiler via property
- Clean up redundant tasks
- Update deps along the way
- Try and fail to use `-xjdk-release`
  - This has confusing behavior with jvmTarget and toolchains
- Remove redundant RequiresOptIn arg
  - Add it back because https://youtrack.jetbrains.com/issue/KT-52720


slack-plugin/src/main/kotlin/slack/gradle/SlackProperties.kt
```diff
@@ -253,6 +253,13 @@ public class SlackProperties private constructor(private val project: Project) {
   public val kotlinDaemonArgs: String
     get() = stringProperty(KOTLIN_DAEMON_ARGS_KEY, defaultValue = "")
 
+  /**
+   * Flag to enable the new K2 compiler. Plumbed into the kotlinOptions.useK2 property and here to
+   * allow for use from a `local.properties` file.
+   */
+  public val useK2: Boolean
+    get() = booleanProperty("slack.useK2", defaultValue = false)
```
seankim-android, line 261: Nice. It looks like there’s a nice performance boost with this new compiler from https://kotlinlang.org/docs/whatsnew17.html#new-kotlin-k2-compiler-for-the-jvm-in-alpha

slack-plugin/src/main/kotlin/slack/gradle/StandardProjectConfigurations.kt
```diff
@@ -815,43 +837,14 @@ internal class StandardProjectConfigurations {
...
-              file("$buildDir/intermediates/detekt/baseline.xml")
-            } else {
-              file(rootProject.file(baselineFile))
-            }
-        }
-      }
-    }
-
     pluginManager.withPlugin("org.jetbrains.kotlin.jvm") {
-      onKotlinPluginApplied()
```
seankim-android, line 845: > Use the new kotlin base plugin as a hook

Question for my understanding. Why did we need to invoke this lambda multiple times before, and we don't need it anymore?

seankim-android (approved):
LGTM

slack-plugin/src/main/kotlin/slack/gradle/StandardProjectConfigurations.kt
ZacSweers: before we had to hook in both the jvm and android plugins, but now we get the "base" plugin as a common hook for these common configs. The android plugin has a similar tool

PR #28, opened 2022-05-20
Predeclare spotless deps instead

ZacSweers:
See https://github.com/diffplug/spotless/issues/1213. Better performance too since we use the same config for every project


seankim-android (approved):
LGTM

PR #31, opened 2022-05-25
Use a custom `ValueSource` to avoid unnecessary configuration cache invalidations

### LINKED ISSUE #30: Git status changes affect configuration cache
We capture these values currently via exec operation since updating to Gradle 7.5, but this breaks configuration cache more often in an undesirable way. We should find a way to ignore this or disable it.

ZacSweers:
Resolves #30, using a custom `ValueSource` per https://docs.gradle.org/7.5-rc-1/release-notes.html#configuration-cache-improvements. The API is a little ugly and explicit but didn't see an easier way to do this.

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/ProcessExtensions.kt
+++ b/slack-plugin/src/main/kotlin/slack/ProcessExtensions.kt
@@ -63,7 +63,7 @@ internal fun executeWithResult(
   arguments: List<String>,
   isRelevantToConfigurationCache: Boolean
 ): Provider<String> {
-  if (isRelevantToConfigurationCache) {
+  if (!isRelevantToConfigurationCache) {
     return providers
       .of(ExecSource::class.java) {
         parameters.workingDir.set(inputWorkingDir)
--- a/slack-plugin/src/main/kotlin/slack/ProcessExtensions.kt
+++ b/slack-plugin/src/main/kotlin/slack/ProcessExtensions.kt
@@ -16,45 +16,75 @@
 package slack
 
 import java.io.File
+import javax.inject.Inject
+import org.gradle.api.file.RegularFileProperty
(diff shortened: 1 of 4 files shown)


ZacSweers:
This still invalidates the configuration cache, so I'm not really sure how to proceed. Will likely need to just remove this data from scans for now until https://github.com/gradle/gradle/issues/20849 is resolved

slack-plugin/src/main/kotlin/slack/ProcessExtensions.kt
```diff
@@ -16,51 +16,123 @@
...
+import org.gradle.process.internal.ExecException
+import slack.ExecSource.Parameters
 import slack.gradle.util.sneakyNull
 
-internal fun String.executeBlocking(providers: ProviderFactory, workingDir: File) {
-  executeBlockingWithResult(providers, workingDir)
+internal fun String.executeBlocking(
+  providers: ProviderFactory,
+  workingDir: File,
+  isRelevantToConfigurationCache: Boolean
```
gavra0, line 37: Most of this should "just work" :) :
- if there are properties read during configuration, then if their value changes, you probably want to re-run, right?
- if there are properties that you need to read, but they aren't really impacting conf (such as build scan tags), that sounds like a missing API

If this provides huge decrease in configuration cache misses, it could be worth merging, but I think it may cause quite a lot of hard-to-debug bugs.

gavra0: That is a very interesting and sad find... :/

Yeah, keeping this as-is for now seems reasonable, but I'd consider filing an issue on your end to keep track of progress on https://github.com/gradle/gradle-org-conventions-plugin/issues/18, and remove this once that lands.

gavra0 (approved):

PR #35, opened 2022-05-28
Update dependencies

ZacSweers:


slack-plugin/src/main/kotlin/slack/gradle/SlackRootPlugin.kt
```diff
@@ -204,29 +196,9 @@ internal class SlackRootPlugin : Plugin<Project> {
...
-            includeGroup("app.cash.sqldelight")
-          }
-          bundle("turbine") {
-            primary("app.cash.turbine:turbine")
-            includeGroup("app.cash.turbine")
-          }
-          bundle("wire") {
-            primary("com.squareup.wire:wire-runtime")
-            includeGroup("com.squareup.wire")
-          }
```
seankim-android, line 229: Were these ever used in the project? 🤔

seankim-android (approved):

slack-plugin/src/main/kotlin/slack/gradle/SlackRootPlugin.kt
ZacSweers: In the android project this applies to yeah, not this one specifically though

PR #39, opened 2022-05-31
Add detekt configs + document auto-applied plugins

ZacSweers:
More work on the reusability front

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/gradle/SlackProperties.kt
+++ b/slack-plugin/src/main/kotlin/slack/gradle/SlackProperties.kt
@@ -332,6 +332,14 @@ public class SlackProperties private constructor(private val project: Project) {
   public val autoApplyCacheFix: Boolean
     get() = booleanProperty("slack.auto-apply.cache-fix", defaultValue = true)
 
+  /* Detekt configs. */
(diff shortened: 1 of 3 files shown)


slack-plugin/src/main/kotlin/slack/gradle/SlackProperties.kt
```diff
@@ -332,6 +332,14 @@ public class SlackProperties private constructor(private val project: Project) {
   public val autoApplyCacheFix: Boolean
     get() = booleanProperty("slack.auto-apply.cache-fix", defaultValue = true)
 
+  /* Detekt configs. */
+  /** Detekt config files, evaluated from rootProject.file(...). */
```
seankim-android, line 336: Wasn't sure if this comment was intended
```suggestion
  /** Detekt config files, evaluated from rootProject.file(...). */
```

seankim-android (approved):

slack-plugin/src/main/kotlin/slack/gradle/SlackProperties.kt
ZacSweers: The first one is a block comment about the multiple below props

PR #40, opened 2022-06-02
Fix spotless applications

ZacSweers:
Formatters need to apparently be decalred in _both_ the subproject and root project, they just must also be the same. Also fixes delete task handling as spotless will apply the base lifecycle plugin for us before-hand now

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/gradle/SlackBasePlugin.kt
+++ b/slack-plugin/src/main/kotlin/slack/gradle/SlackBasePlugin.kt
@@ -55,10 +55,6 @@ internal class SlackBasePlugin : Plugin<Project> {
         target.apply(plugin = "org.gradle.test-retry")
       }
 
-      if (slackProperties.autoApplySpotless) {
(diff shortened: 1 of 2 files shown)


ZacSweers:
github actions seems to be having issues

slack-plugin/src/main/kotlin/slack/gradle/SlackBasePlugin.kt
```diff
@@ -117,64 +113,76 @@ internal class SlackBasePlugin : Plugin<Project> {
...
-            target("*.md", ".gitignore")
-            trimTrailingWhitespace()
-            endWithNewline()
-          }
+      val spotlessFormatters: SpotlessExtension.() -> Unit = {
+        format("misc") {
+          target("*.md", ".gitignore")
+          trimTrailingWhitespace()
+          endWithNewline()
+        }
```
seankim-android, line 127: Nice!

seankim-android (approved):

PR #68, opened 2022-07-30
Update OkHttp to 5.0.0-alpha.10

ZacSweers:


seankim-android (approved):
I thought we moved away from alpha because of some reason... Looks like we were on it the whole time!

PR #85, opened 2022-08-12
Don't skip on empty sources in LocTask

ZacSweers:
We want to always run this because not every module has source files. In those cases we want to just write an empty JSON object for tasks that depend on this task's outputs. This fixes an issue we would see in resources-only projects


seankim-android (approved):

seankim-android:
Looks good to me, but the CI checks are failing

PR #108, opened 2022-09-06
Implement compose() feature DSL in SlackExtension

ZacSweers:
This moves compose support to a new DSL configuration instead.

```kotlin
slack {
  android {
    features {
      compose()
    }
  }
}
```

Currently there is no extra configuration beyond the above in the DSL, but we could add more in the future. This also adds a new `defaultComposeAndroidBundleAlias` property that can be used to identify a default `libs.bundle.*` alias to include in any projects that enable compose. This is, in this case, just focused on adding _core_ compose deps (runtime, ui, etc).

One implementation note is that due to AGP using regular kotlin properties still, we have to make enabling compose a synchronous action that directly sets them on the underlying AGP property under the hood. If they ever convert them into lazy gradle properties, we could look at chaining these instead.

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/gradle/SlackExtension.kt
+++ b/slack-plugin/src/main/kotlin/slack/gradle/SlackExtension.kt
@@ -17,12 +17,14 @@
 
 package slack.gradle
 
+import com.android.build.api.dsl.CommonExtension
 import com.squareup.anvil.plugin.AnvilExtension
 import dev.zacsweers.moshix.ir.gradle.MoshiPluginExtension
 import javax.inject.Inject
 import org.gradle.api.Action
 import org.gradle.api.Project
 import org.gradle.api.artifacts.ProjectDependency
+import org.gradle.api.artifacts.VersionCatalog
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Property
 import org.gradle.kotlin.dsl.apply
@@ -38,9 +40,16 @@ import slack.gradle.dependencies.SlackDependencies
 @DslMarker public annotation class SlackExtensionMarker
 
 @SlackExtensionMarker
-public abstract class SlackExtension @Inject constructor(objects: ObjectFactory) {
-
-  internal val androidHandler = objects.newInstance<AndroidHandler>()
+public abstract class SlackExtension
+@Inject
+constructor(
+  objects: ObjectFactory,
+  globalSlackProperties: SlackProperties,
+  private val slackProperties: SlackProperties,
+  versionCatalog: VersionCatalog
+) {
+  internal val androidHandler =
+    objects.newInstance<AndroidHandler>(globalSlackProperties, slackProperties, versionCatalog)
   internal val featuresHandler = objects.newInstance<FeaturesHandler>()
 
   public fun android(action: Action<AndroidHandler>) {
@@ -51,7 +60,7 @@ public abstract class SlackExtension @Inject constructor(objects: ObjectFactory)
     action.execute(featuresHandler)
   }
 
-  internal fun configureFeatures(project: Project, slackProperties: SlackProperties) {
+  internal fun applyTo(project: Project) {
     val logVerbose = slackProperties.slackExtensionVerbose
     // Dirty but necessary since the extension isn't configured yet when we call this
     project.afterEvaluate {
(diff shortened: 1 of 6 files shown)


slack-plugin/src/main/kotlin/slack/gradle/SlackExtension.kt
```diff
@@ -38,9 +40,16 @@ import slack.gradle.dependencies.SlackDependencies
...
 @SlackExtensionMarker
-public abstract class SlackExtension @Inject constructor(objects: ObjectFactory) {
-
-  internal val androidHandler = objects.newInstance<AndroidHandler>()
+public abstract class SlackExtension
+@Inject
+constructor(
+  objects: ObjectFactory,
+  globalSlackProperties: SlackProperties,
+  private val slackProperties: SlackProperties,
```
autonomousapps, line 48: For the uninitiated (me), where are there two instances of the type `SlackProperties`?

slack-plugin/src/main/kotlin/slack/gradle/SlackExtension.kt
```diff
@@ -636,12 +645,73 @@ public abstract class DaggerHandler @Inject constructor(objects: ObjectFactory)
...
+  internal val enabled = objects.property<Boolean>().convention(false)
+
+  /**
+   * This is weird! Due to the non-property nature of the AGP buildFeatures and composeOptions DSLs,
+   * we can't lazily chain their values to our own extension's properties. Because of this, we
+   * lazily set this instance from [StandardProjectConfigurations] during Android extension
+   * evaluation and then make calls to [enable] directly set the values on this instance. Ideally we
+   * could eventually remove this if/when AGP finally makes these properties lazy.
+   */
+  internal var androidExtension: CommonExtension<*, *, *, *>? = null
```
autonomousapps, line 674: I have similar-ish code in Square's AGP anti-corruption layer (facade/wrapper). We actually instantiate the facade lazily and pass it in via the `@Inject`ed constructor.

slack-plugin/src/main/kotlin/slack/gradle/SlackExtension.kt
```diff
@@ -636,12 +645,73 @@ public abstract class DaggerHandler @Inject constructor(objects: ObjectFactory)
...
+
+  internal fun enable() {
+    val extension =
+      checkNotNull(androidExtension) {
+        "ComposeHandler must be configured with an Android extension before it can be enabled. Did you apply the Android gradle plugin?"
+      }
+    enabled.set(true)
```
autonomousapps, line 681: One thing I like to do after calling this is to then call `enabled.disallowChanges()`. That way, I know that the value can only ever be set once. Probably less relevant in this case, since it can only be "enabled", but we have a lot of true|false settings.

autonomousapps: Unrelated, and this is a matter of taste more than anything I think, but in some cases I've delegated the call to a companion object on a specific plugin, solely so that I could "encapsulate" all related logic in one place. E.g., we have a "UnitTestPlugin", and so we'll call static methods on that from a Handler class like this, again just to keep all test-related logic in one place for discoverability.

autonomousapps (approved):

slack-plugin/src/main/kotlin/slack/gradle/SlackExtension.kt
ZacSweers: One is global properties (I.e. root project) and the other is the local project. It's a bit of a holdover from before I built our new properties system though so I plan to consolidate them, as the new property API I have will gracefully fallback to the global one if the project doesn't specify

slack-plugin/src/main/kotlin/slack/gradle/SlackExtension.kt
ZacSweers: Nice. I couldn't find an easy way to make it constructor injected in our side for now

slack-plugin/src/main/kotlin/slack/gradle/SlackExtension.kt
ZacSweers: Could you unpack that a bit more for me?

ZacSweers: disallow changes here [9d5a5c9](https://github.com/slackhq/slack-gradle-plugin/pull/108/commits/9d5a5c9bae9c4eb4464d1506a5eea6815bd6f9aa). Probably more places we should be doing this tbh

PR #111, opened 2022-09-06
Prepare for publishing on central

ZacSweers:
This sets up publishing for maven central.

- New group ID and artifact coordinates to differentiate from the previous versions.
- We now track a real changelog.
- Publish snapshots on the main branch.
- No longer support publishing to our own artifactory. We may add this support back in the future, but not seeing much of a point currently since the time delay is ~20min these days and we can test with snapshots first.

I'd like to also set up easy publishing for other slack folks in subsequent steps via in-memory key, but continuing manually for now should be fine


RELEASING.md
```diff
@@ -0,0 +1,12 @@
+Releasing
+=========
+
+1. Change the version in `gradle.properties` to a non-SNAPSHOT version.
+2. Update the `CHANGELOG.md` for the impending release.
+3. `git commit -am "Prepare for release X.Y.Z."` (where X.Y.Z is the new version)
+4. `git tag -a X.Y.Z -m "Version X.Y.Z"` (where X.Y.Z is the new version)
+5. `./gradlew clean publish --no-daemon --no-parallel && ./gradlew closeAndReleaseRepository`
```
autonomousapps, line 8: `clean --no-daemon --no-parallel` why you do this to me

autonomousapps (approved):

PR #115, opened 2022-09-08
Specify kotlin version in compose compatibility check

ZacSweers:
This is an annoying new granularity to the compose compiler version checks that we have to handle now.

kotlin 1.7.10, compose expects 1.7.0 => specify

1.7.10 kotlin 1.7.10, compose expects 1.7.10 => omit the arg entirely

https://android-review.googlesource.com/c/platform/frameworks/support/+/2189520


seankim-android (approved):
🚀

PR #125, opened 2022-09-27
Set lint config in android and JVM libraries too

ZacSweers:
We were omitting this before unnecessarily, and it prevents these from reusing shared configuration


kierse (approved):
looks good 👍

PR #126, opened 2022-09-27
Two quick androidTest allowed variants fixes

ZacSweers:
- Fix using the wrong name for the flavor name check
- Fix not applying this in application projects too


kierse (approved):
Looks good

PR #132, opened 2022-10-03
Add best practices plugin + two fixes

ZacSweers:
Integrates https://github.com/autonomousapps/gradle-best-practices-plugin

Knew about these two issues before, but was a good forcing function to properly address them

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/gradle/SlackBasePlugin.kt
+++ b/slack-plugin/src/main/kotlin/slack/gradle/SlackBasePlugin.kt
@@ -29,9 +29,12 @@ import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.testing.Test
 import org.gradle.kotlin.dsl.apply
 import org.gradle.kotlin.dsl.configure
+import org.gradle.kotlin.dsl.named
 import org.gradle.kotlin.dsl.retry
 import org.gradle.kotlin.dsl.withType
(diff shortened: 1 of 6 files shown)


slack-plugin/src/main/kotlin/slack/gradle/SlackBasePlugin.kt
```diff
@@ -347,4 +351,20 @@ internal class SlackBasePlugin : Plugin<Project> {
       }
     }
   }
+
+  private fun Project.configureBootstrapTasks() {
+    val rootTask = rootProject.tasks.named<CoreBootstrapTask>(CoreBootstrapTask.NAME)
```
autonomousapps, line 356: fyi at some point, I expect gradle _may_ make `rootProject` access "bad" too. I'm hoping they'll make some effort to provide a less-than-friction-full path, though.

slack-plugin/src/main/kotlin/slack/stats/ModuleStats.kt
```diff
@@ -72,19 +72,9 @@ public object ModuleStatsTasks {
   internal fun configureRoot(rootProject: Project) {
     val includeGenerated = rootProject.includeGenerated()
 
-    val projectAccessorMapping =
-      rootProject.provider {
-        rootProject.subprojects.associate {
```
autonomousapps, line 77: a-ha, `subprojects`, my old enemy

autonomousapps (approved):

PR #133, opened 2022-10-03
Only configure bootstrap conditionally

ZacSweers:
It may not always be enabled

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/gradle/tasks/BootstrapTask.kt
+++ b/slack-plugin/src/main/kotlin/slack/gradle/tasks/BootstrapTask.kt
@@ -48,6 +48,7 @@ import org.gradle.jvm.toolchain.JavaLanguageVersion
 import org.gradle.jvm.toolchain.JavaLauncher
 import org.gradle.jvm.toolchain.JavaToolchainService
 import org.gradle.kotlin.dsl.listProperty
+import org.gradle.kotlin.dsl.named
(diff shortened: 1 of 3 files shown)


slack-plugin/src/main/kotlin/slack/gradle/tasks/BootstrapTask.kt
```diff
@@ -308,13 +309,34 @@ constructor(objects: ObjectFactory, providers: ProviderFactory) : DefaultTask()
...
-    public const val NAME: String = "bootstrap"
+    private const val NAME: String = "bootstrap"
 
     /**
      * Current bootstrap version. Other bootstrap tasks can check against this as a minimum version.
      */
     public const val VERSION: Int = 2
 
+    internal fun isBootstrapEnabled(project: Project): Boolean {
+      return project.gradle.startParameter.taskNames.any { it == "bootstrap" }
```
kierse, line 320: Looks like you could use `NAME` here (and on line 330 as well)

ZacSweers: good point, I'll fix that

kierse (approved):

slack-plugin/src/main/kotlin/slack/gradle/tasks/BootstrapTask.kt
ZacSweers: 789a6d6

PR #137, opened 2022-10-06
Only check allowed androidTest variants if any are defined

ZacSweers:
We initially set this up to default to treating all variants as disabled if none were defined, when we should be doing the opposite.


slack-plugin/src/main/kotlin/slack/gradle/StandardProjectConfigurations.kt
```diff
@@ -645,9 +645,9 @@ internal class StandardProjectConfigurations(
             slackExtension.androidHandler.featuresHandler.androidTest.getOrElse(false)
           val variantEnabled =
             androidTestEnabled &&
-              builder.name in
-                slackExtension.androidHandler.featuresHandler.androidTestAllowedVariants.orNull
-                  .orEmpty()
+              slackExtension.androidHandler.featuresHandler.androidTestAllowedVariants.orNull
```
hellohj, line 648: Is the comment from line 642 still relevant after this change?

valeraz (approved):

PR #139, opened 2022-10-11
Automatically add compose compiler dep in multiplatform projects

ZacSweers:
This forces the project-defined compose compiler version, which is often newer than the actual compose-jb versions.


slack-plugin/src/main/kotlin/slack/gradle/SlackExtension.kt
```diff
@@ -633,6 +634,14 @@ constructor(
   internal fun applyTo(project: Project) {
     if (enabled.getOrElse(false)) {
       composeBundleAlias?.let { project.dependencies.add("implementation", it) }
+      project.pluginManager.withPlugin("org.jetbrains.kotlin.multiplatform") {
```
ZacSweers, line 637:
```suggestion
      project.pluginManager.withPlugin("org.jetbrains.compose") {
```

adamsp (approved):

PR #176, opened 2022-12-03
Only enable androidResources in tests if robolectric is requested

ZacSweers:
MergeResources is an expensive task and we currently enable it on every android project by default. However, only Robolectric needs this in our needs, so we can save time by only enabling it when tests need it.

In our internal codebase this brings our CI unit test job down from ~18k tasks to ~15k total tasks. Working on benchmarks with this next to see time improvements.

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/gradle/SlackProperties.kt
+++ b/slack-plugin/src/main/kotlin/slack/gradle/SlackProperties.kt
@@ -370,6 +370,12 @@ public class SlackProperties private constructor(private val project: Project) {
   public val strictValidateAndroidTestManifest: Boolean
     get() = booleanProperty("slack.strict.validateAndroidTestManifests", defaultValue = true)
 
+  /**
+   * Always enables resources in android unit tests. Only present for benchmarking purposes and should otherwise be off.
+   */
+  public val alwaysEnableResourcesInTests: Boolean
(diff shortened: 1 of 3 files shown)


slack-plugin/src/main/kotlin/slack/gradle/SlackProperties.kt
```diff
@@ -370,6 +370,13 @@ public class SlackProperties private constructor(private val project: Project) {
   public val strictValidateAndroidTestManifest: Boolean
     get() = booleanProperty("slack.strict.validateAndroidTestManifests", defaultValue = true)
 
+  /**
+   * Always enables resources in android unit tests. Only present for benchmarking purposes and
+   * should otherwise be off.
+   */
+  public val alwaysEnableResourcesInTests: Boolean
+    get() = booleanProperty("slack.gradle.config.test.alwaysEnableResources", defaultValue = false)
```
autonomousapps, line 378: If it should only be enabled for benchmarking, do you think it would make sense to emit a warning when it's enabled?

autonomousapps (approved):
LGTM!

slack-plugin/src/main/kotlin/slack/gradle/SlackProperties.kt
ZacSweers: I plan to remove it basically as soon as we're done benchmarking ¯\_(ツ)_/¯

PR #187, opened 2022-12-10
Backport new jniLibs access in AGP 8 handling

ZacSweers:
The returned class went from `JniPackagingOptions` to `JniPackaging`, which is a signature change that breaks at runtime.


autonomousapps (approved):
LGTM.

PR #193, opened 2022-12-17
Don't apply KotlinCompile configs automatically to stub gen tasks

ZacSweers:
KGP automatically copies arguments from the "real" kotlincompile task [here](https://github.com/JetBrains/kotlin/blob/0e4e53786c1b0341befe8e71a5e6e0bc0e464370/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/internal/kapt/KaptGenerateStubsTask.kt#L111-L113) and can result in duplicates if we add them on our own.


ZacSweers:
surprise!

gabrielittner:
@ZacSweers did you file an issue for this?

ZacSweers:
https://youtrack.jetbrains.com/issue/KT-55565

gabrielittner:
Thanks 👍

PR #203, opened 2023-01-03
Implement SgpSettingsPlugin

ZacSweers:
This implements a new `Settings` plugin in Gradle for global configuration of different sorts.

Highlights
- Auto-configure AGP 8.x's new `Settings` plugin with values for `minSdk` and `compileSdk` + expose settings hook in `AgpHandlerFactory` as a toe-hold for the future.
- Expose a DSL for configuring `SlackVersions` values, defaulting to those in the `libs.versions.toml` file.
- Cache `AgpHandler` from the settings plugin via depositing the instance in project extras, borrowing from what AGP does.
- Automatically apply SGP's root and base plugins across all subprojects (!!)

Breaking change: minSdk and compileSdk must now move to `libs.versions.toml`.

Example
```kotlin
plugins {
  id("com.slack.gradle.settings")
}

slack {
  versions {
    jvmTarget.set(11)
  }
}
```

TODO in future PRs
- Move more props from `GlobalConfig` to this.

### DIFF
--- a/agp-handlers/agp-handler-api/src/main/kotlin/slack/gradle/agp/AgpHandlerFactory.kt
+++ b/agp-handlers/agp-handler-api/src/main/kotlin/slack/gradle/agp/AgpHandlerFactory.kt
@@ -15,6 +15,8 @@
  */
 package slack.gradle.agp
 
+import java.util.ServiceLoader
+
 /**
  * A basic factory interface for creating [AgpHandler] instances. These should be implemented and
  * contributed as a service loader via something like `@AutoService`.
@@ -30,10 +32,36 @@ public interface AgpHandlerFactory {
    * Creates a new [AgpHandler] instance for the current AGP version if [currentVersion] was deemed
    * to match.
    */
-  public fun create(): AgpHandler
+  public fun createHandler(): AgpHandler
   /**
    * Creates a new [AgpSettingsHandler] instance for the current AGP version if [currentVersion] was
    * deemed to match.
    */
   public fun createSettingsHandler(): AgpSettingsHandler = AgpSettingsHandler.NoOp
+
+  public companion object {
+    public fun load(): AgpHandlerFactory {
+      /**
+       * Load handlers and pick the highest compatible version (by [AgpHandlerFactory.minVersion])
+       */
+      return ServiceLoader.load(AgpHandlerFactory::class.java)
+        .iterator()
+        .asSequence()
+        .mapNotNull { factory ->
+          // Filter out any factories that can't compute the AGP version, as
+          // they're _definitely_ not compatible
+          try {
+            FactoryData(VersionNumber.parse(factory.currentVersion()), factory)
+          } catch (t: Throwable) {
+            null
+          }
+        }
+        .filter { (agpVersion, factory) -> agpVersion.baseVersion >= factory.minVersion }
+        .maxByOrNull { (_, factory) -> factory.minVersion }
+        ?.factory
+        ?: error("Unrecognized AGP version!")
+    }
+  }
 }
+
+private data class FactoryData(val agpVersion: VersionNumber, val factory: AgpHandlerFactory)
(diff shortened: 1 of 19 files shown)


agp-handlers/agp-handler-api/src/main/kotlin/slack/gradle/agp/AgpHandlerFactory.kt
```diff
@@ -24,7 +26,42 @@ package slack.gradle.agp
...
+          try {
+            FactoryData(VersionNumber.parse(factory.currentVersion()), factory)
+          } catch (t: Throwable) {
+            null
+          }
+        }
+        .filter { (agpVersion, factory) -> agpVersion.baseVersion >= factory.minVersion }
+        .maxByOrNull { (_, factory) -> factory.minVersion }
+        ?.factory
+        ?: error("Unrecognized AGP version!")
```
autonomousapps, line 62: Is this good enough, or can the error message contain a more actionable message?

agp-handlers/agp-handler-api/src/main/kotlin/slack/gradle/agp/AgpHandlerFactory.kt
```diff
@@ -24,7 +26,42 @@ package slack.gradle.agp
...
+   * deemed to match.
+   */
+  public fun createSettingsHandler(): AgpSettingsHandler = AgpSettingsHandler.NoOp
+
+  public companion object {
+    public fun load(): AgpHandlerFactory {
+      /**
+       * Load handlers and pick the highest compatible version (by [AgpHandlerFactory.minVersion])
+       */
+      return ServiceLoader.load(AgpHandlerFactory::class.java)
```
autonomousapps, line 47: Interesting approach.

slack-plugin/src/main/kotlin/slack/gradle/settings/SgpSettingsPlugin.kt
```diff
@@ -0,0 +1,174 @@
...
+    // deposit the handler instance into each project using the extra Properties.
+    settings.gradle.beforeProject {
+      // Gradle does this immensely weird thing where it considers all intermediary folders in
+      // between projects to _also_ be projects. We skip those in configuration.
+      if (!buildFile.exists()) return@beforeProject
+
+      extensions.extraProperties[SGP_VERSIONS_EXTENSION_KEY] = versions
+      extensions.extraProperties[AGP_HANDLER_EXTENSION_KEY] = handler
+
+      // Apply base plugin to all subprojects
```
autonomousapps, line 64: I'm curious about how this works in practice for you. In our set of convention plugins, we also have a base plugin (because of course), but we also have baked in the assumption that the base plugin is applied first, and then subsequent plugins can then count on that being true. I assume that `beforeProject {}` ensures the base plugin is applied first, right? And just thinking out loud... I guess the fact that these plugins are applied via the settings script means they're loaded in the settings classloader, whereas other plugins might be loaded in different (probably child) classloaders. I guess everything "just works"? But I always wonder a bit with gradle's crazy classloader hierarchy.

autonomousapps (approved):

agp-handlers/agp-handler-api/src/main/kotlin/slack/gradle/agp/AgpHandlerFactory.kt
ZacSweers: Good enough for now, it's what we had before too

slack-plugin/src/main/kotlin/slack/gradle/settings/SgpSettingsPlugin.kt
ZacSweers: SGP is more or less order-agnostic when it comes to other plugins, so this wouldn't change anything for us at least!

ZacSweers:
huh, I actually have no idea how to use a settings plugin from an included build

PR #216, opened 2023-01-17
Update to JDK 19 and latest AGPs

ZacSweers:


agp-handlers/agp-handler-74/src/main/kotlin/slack/gradle/agphandler/v74/AgpHandlerFactory73.kt
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package slack.gradle.agphandler.v73
+package slack.gradle.agphandler.v74
```
kierse, line 16: Minor: should this file be renamed? Or drop 73 from the name? I see that it now lives in a version specific folder...

kierse (approved):

agp-handlers/agp-handler-74/src/main/kotlin/slack/gradle/agphandler/v74/AgpHandlerFactory73.kt
ZacSweers: Oops good catch, will fix

PR #224, opened 2023-01-23
Introduce monkey patch for MergeFileTask

ZacSweers:
After some investigation, I believe this task is currently breaking remote caching for lint because it has non-deterministic ordering of file merges, resulting in the merged file's content to be different. This introduces a monkey patch of the task (that is - we'll put it on the classpath first so that AGP uses our implementation) that has a system property guard to enable sorting of these files first. This allows us to test this theory out. I've also created a standalone monkeypatch artifact as a toe hold for any future ones we want to add too.

Note the task is copied identically from AGP sources with the exception of the extra block for filtering files.

### DIFF
--- a/sgp-monkeypatch-agp/src/main/kotlin/com/android/build/gradle/internal/tasks/MergeFilesTask.kt
+++ b/sgp-monkeypatch-agp/src/main/kotlin/com/android/build/gradle/internal/tasks/MergeFilesTask.kt
@@ -1,0 +1,94 @@
+/*
+ * Copyright (C) 2013 Slack Technologies, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


sgp-monkeypatch-agp/src/main/kotlin/com/android/build/gradle/internal/tasks/MergeFilesTask.kt
```diff
@@ -0,0 +1,91 @@
...
+    public fun mergeFiles(inputFiles: Collection<File>, output: File) {
+      // filter out any non-existent files
+      val existingFiles = inputFiles.filter { it.isFile() }
+
+      val filesToUse =
+        if (System.getenv("com.slack.sgp.sort-merge-files").toBoolean()) {
+          existingFiles.sortedBy { it.name.lowercase(Locale.US) }
+        } else {
+          existingFiles
+        }
```
ZacSweers, line 69: This is our sorting change

valeraz (approved):

PR #227, opened 2023-01-26
Add more logging to monkeypatch and also sort by absolute path

ZacSweers:
Trying to debug some things as the previous version isn't working

### DIFF
--- a/sgp-monkeypatch-agp/src/main/kotlin/com/android/build/gradle/internal/tasks/MergeFilesTask.kt
+++ b/sgp-monkeypatch-agp/src/main/kotlin/com/android/build/gradle/internal/tasks/MergeFilesTask.kt
@@ -63,7 +63,10 @@ public abstract class MergeFileTask : NonIncrementalTask() {
 
       val filesToUse =
         if (System.getenv("com.slack.sgp.sort-merge-files").toBoolean()) {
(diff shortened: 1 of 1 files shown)


sgp-monkeypatch-agp/src/main/kotlin/com/android/build/gradle/internal/tasks/MergeFilesTask.kt
```diff
@@ -63,7 +63,10 @@ public abstract class MergeFileTask : NonIncrementalTask() {
 
       val filesToUse =
         if (System.getenv("com.slack.sgp.sort-merge-files").toBoolean()) {
-          existingFiles.sortedBy { it.name.lowercase(Locale.US) }
+          if (System.getenv("com.slack.sgp.sort-merge-files.log").toBoolean()) {
+            println("Sorting ${existingFiles.size} files in merge task. Files are $existingFiles")
+          }
+          existingFiles.sortedBy { it.absolutePath.lowercase(Locale.US) }
```
gavra0, line 69: To ensure consistency across OS(s) you can use `it. invariantSeparatorsPath`.

ZacSweers: ooo TIL

gavra0 (approved):

PR #256, opened 2023-02-16
Set up git hooks and offer them in SGP too

### LINKED ISSUE #255: Allow bootstrap to write and prepare common git hooks

ZacSweers:
This upstreams our standard git hooks (with some internal things removed) so that any project using SGP can use them to hit the ground running.

The entry point is `./gradlew installCommitHooks`, which will install them to `config/git/hooks` and then print a message saying how to install them.

The intent is for this to be a starting point for kotlin and java formatting, and then projects can add their own custom configurations on top of it (for example - detekt).

This also sets up git hooks and formatting binaries in this project itself, and uses them as the template that this task uses (they're copied into resources).

Resolves #255

### DIFF
--- a/config/git/hooks/post-checkout
+++ b/config/git/hooks/post-checkout
@@ -1,0 +1,3 @@
+#!/bin/sh
+command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting '.git/hooks/post-checkout'.\n"; exit 2; }
+git lfs post-checkout "$@"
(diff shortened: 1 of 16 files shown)


config/git/hooks/post-checkout
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting '.git/hooks/post-checkout'.\n"; exit 2; }
```
adamsp, line 2: I think these comments have the wrong path - we install to `<root>/config/git/hooks/` but these reference `.git/hooks/`

ZacSweers: good point, we should fix this in our internal ones too heh

ZacSweers: Fixed!

ZacSweers: [773509a](https://github.com/slackhq/slack-gradle-plugin/pull/256/commits/773509a9fadf221e84f17cd6fbe68ea6d07c024f)

adamsp (approved):

PR #277, opened 2023-03-07
Revert "Add lintErrorRuleIds property"

ZacSweers:
Reverts slackhq/slack-gradle-plugin#276. I completely forgot that lint.xml is a thing and that's the right solution for this case.


adamsp (approved):
Oops

PR #280, opened 2023-03-10
Revamp unit test handling

ZacSweers:
- Introduce `globalCiUnitTest` lifecycle task for convenience
- Remove unit test plugin, make it just a part of standard project configuration. Long overdue!
- Clean up and consolidate `Test` task configurations.
- Configure `Test` tasks the same both locally and on CI. I suspect this is part of what was negatively affecting caching for these tasks. A bunch was left over from past work around memory issues, and I've batched those together behind the existing verbose logging flag for whenever we need to investigate those.


ZacSweers:
Tested on the internal slack repo and confirmed `ciUnitTest --dry-run | grep ciUnitTest` and `globalCiUnitTest --dry-run | grep ciUnitTest` yield the same number of tasks 👍

slack-plugin/src/main/kotlin/slack/unittest/UnitTests.kt
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * For convenience, this also creates a task named `compileCiUnitTest` for just _building_ the
+ * tests.
+ *
+ * Finally, this plugin creates a convenience task named "globalCiUnitTest" in the root project,
+ * which depends on all the "ciUnitTest" tasks in the subprojects. This is the task that should be
+ * run on CI.
+ */
+internal object UnitTests {
+  private val MAX_PARALLEL = max(Runtime.getRuntime().availableProcessors() / 2, 1)
```
valeraz, line 53: Curious if giving more CPUs to the task is something that we want to experiment with.

valeraz (approved):

slack-plugin/src/main/kotlin/slack/unittest/UnitTests.kt
ZacSweers: We could yeah. I'll add an override we could test

ZacSweers: [b7c2ea9](https://github.com/slackhq/slack-gradle-plugin/pull/280/commits/b7c2ea9fb81666a4b232eb3ae85cce0bea7b1a20)

PR #286, opened 2023-03-15
Implement ComputeAffectedProjects

ZacSweers:
This implements a new `ComputeAffectedProjects` task. Most of the relevant docs are on the class and its properties/function itself, but the overview is this task is that it's used to compute affected projects from a given input of changed files (i.e. from a pull request).

**Logic Flow**
1. Parse input changed files
2. Filter input files against a set of input glob patterns to include only relevant files (i.e. source files like `.kt` files)
    - If no files patch, no projects are affected!
3. Check filtered files against a set of "never-skippable" glob patterns, such as version catalog files or root build files. If any file matches, the whole change is deemed not skippable and no file is emitted. This, in practice, means "all projects are affected".
4. Map filtered changed files to changed projects
5. Given a set of changed projects, inspect the build graph (reusing some logic from the graph-assert plugin) and determine affected projects based on dependent projects of the changed projects.
6. Output an affected_projects.txt file of affected projects

**Extras**
- This also outputs a `focus.settings.gradle` file, which can be used with https://github.com/dropbox/focus to only load the projects needed to run affected projects. This further reduces overhead.
- There's a lot of helpful diagnostics and debug logging available via the new `slack.debug` property.
- The computation task is _fast_, runs on the order of milliseconds.
- If a project's changed files are determined to be only tests or the lint baseline, it does _not_ mark dependents as affected. A nice little extra optimization.

This does _not_ implement integration into unit tests, lint tasks, and android test APKs yet. Saving that for a separate PR to simplify this one. I have been testing it extensively locally with the Slack android codebase using those integrations and debug diagnostics though!

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/gradle/avoidance/ComputeAffectedProjects.kt
+++ b/slack-plugin/src/main/kotlin/slack/gradle/avoidance/ComputeAffectedProjects.kt
@@ -158,7 +158,14 @@ internal abstract class ComputeAffectedProjects : DefaultTask() {
             gradlePath to ChangedProject(projectPath, gradlePath, files.toSet())
           }
       }
-    log("changedProjects: $changedProjects")
+    writeDiagnostic("changedProjects.txt") {
+      changedProjects.entries
+        .sortedBy { it.key }
+        .joinToString("\n") { (_, v) ->
+          val testOnlyString = if (v.onlyTestsAreChanged) " (test only)" else ""
+          "${v.gradlePath}$testOnlyString\n${v.changedPaths.sorted().joinToString("\n") { "-- $it" } }"
+        }
+    }
 
     val graph = logTimedValue("creating graph") { DependencyGraph.create(dependencyGraph.get()) }
 
@@ -205,6 +212,7 @@ internal abstract class ComputeAffectedProjects : DefaultTask() {
       }
     }
 
+    logger.lifecycle("$LOG Found ${allAffectedProjects.size} affected projects.")
     outputFile.get().asFile.writeText(allAffectedProjects.sorted().joinToString("\n"))
   }
 
@@ -233,7 +241,9 @@ internal abstract class ComputeAffectedProjects : DefaultTask() {
 
     private val DEFAULT_NEVER_SKIP_PATTERNS =
       listOf(
+        // root build.gradle.kts and settings.gradle.kts files
         "*.gradle.kts",
+        // root gradle.properties file
         "gradle.properties",
         "**/*.versions.toml",
         "ci/**",
@@ -245,6 +255,7 @@ internal abstract class ComputeAffectedProjects : DefaultTask() {
       slackProperties: SlackProperties
     ): TaskProvider<ComputeAffectedProjects> {
       // TODO any others?
+      // TODO what about testFixtures?
       val configurationsToLook: Set<String> =
         setOf(
           "api",
@@ -252,7 +263,8 @@ internal abstract class ComputeAffectedProjects : DefaultTask() {
           "ksp",
           "testImplementation",
           "androidTestImplementation",
-          "compileOnly"
+          "compileOnly",
+          "annotationProcessor",
         )
 
       val moduleGraph by lazy {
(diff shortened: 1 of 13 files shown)


ZacSweers:
Example run

image

slack-plugin/src/main/kotlin/slack/gradle/avoidance/ComputeAffectedProjects.kt
```diff
@@ -0,0 +1,520 @@
...
+ * also output timings.
+ *
+ * ### Usage
+ * Example usage:
+ * ```bash
+ * ./gradlew computeAffectedProjects --changed-files changed_files.txt
+ * ```
+ */
+@UntrackedTask(because = "This task is a meta task that more or less runs as a utility script.")
+internal abstract class ComputeAffectedProjects : DefaultTask() {
```
valeraz, line 92: Is there a way to add some unit tests for this wonderful code?

slack-plugin/src/main/kotlin/slack/gradle/avoidance/ComputeAffectedProjects.kt
```diff
@@ -0,0 +1,520 @@
...
+    private const val LOG = "[Skippy]"
+
+    private val DEFAULT_INCLUDE_PATTERNS =
+      listOf(
+        "**/*.kt",
+        "*.gradle.kts",
+        "**/*.gradle.kts",
+        "**/*.java",
+        "**/*.xml",
+        "**/gradle.properties",
```
valeraz, line 343: what about assets... like .png files and such? Could they have an impact on screenshots tests?

valeraz, line 343: Also `*.json` in files like https://github.com/tinyspeck/slack-android-ng/blob/main/features/huddles/src/main/res/raw/huddle_effect_clap.json?

valeraz (commented):
Overall, looks good. My main concern is about how to test this thoroughly to make sure we're not missing anything (e.g. setting the right filters). Seems like we should run it in debug mode on PRs for a while and manually examine the output to make sure it's doing the right thing.

slack-plugin/src/main/kotlin/slack/gradle/avoidance/ComputeAffectedProjects.kt
ZacSweers: hmm, good point. This is just the default set, so I'll add this but for `**/res/**/*.png` only. For the other ones I think we should configure them manually in the slack repo when we integrate it as it's more specific to us.

ZacSweers: see https://github.com/slackhq/slack-gradle-plugin/pull/286#discussion_r1139488771

slack-plugin/src/main/kotlin/slack/gradle/avoidance/ComputeAffectedProjects.kt
```diff
@@ -0,0 +1,520 @@
...
+
+    private val DEFAULT_NEVER_SKIP_PATTERNS =
+      listOf(
+        // root build.gradle.kts and settings.gradle.kts files
+        "*.gradle.kts",
+        // root gradle.properties file
+        "gradle.properties",
+        "**/*.versions.toml",
+        "ci/**",
+        ".buildkite/**",
```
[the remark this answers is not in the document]
ZacSweers: see https://github.com/slackhq/slack-gradle-plugin/pull/286#discussion_r1139488771

ZacSweers:
Heh, yeah I've been thinking a bunch on tests and not quite sure yet what the right tack is. Most of this could probably be abstracted into a standalone program and just make gradle a wrapper around it. Another alternative is to make full blown gradle test suite with stub projects.

What I'd like to do to start for now is run with this in parallel and work with this initial implementation upfront. This way we can start letting this bake and get the other moving parts (metrics, parallel CI jobs, etc) and I can take extra time to make a fuller test suite. lmk what you think

ZacSweers:
Ok did a bunch of initial ground work to make this more testable by separating out the underlying program and basing it on top of Okio's FileSystem APIs. I've added a couple smoke tests for now, but would like to continue with these in parallel.

valeraz (approved):

ZacSweers:
whoops, I forgot to actually push those changes

PR #298, opened 2023-03-21
Fix Detekt build caching

ZacSweers:
Works around https://github.com/detekt/detekt/issues/5925


ZacSweers:
@valeraz reviewed and approved

PR #331, opened 2023-04-25
Don't enable kover on the platform project

ZacSweers:
These don't have unit tests


jstewart5000 (approved):
Do we have to explicitly exclude all projects that don't have unit tests? Thought this would of been somehow figured out automagically by the `kover` plugin

PR #332, opened 2023-04-26
Gracefully handle extensions that fail to load

ZacSweers:
Better to error about them than fail the build


slack-plugin/src/main/kotlin/slack/gradle/SlackTools.kt
```diff
@@ -94,7 +94,14 @@ public abstract class SlackTools : BuildService<Parameters>, AutoCloseable {
         .stream()
         .asSequence()
         .forEach { provider ->
-          val previous = put(provider.type(), provider.get())
+          val extension =
+            try {
+              provider.get()
+            } catch (e: InstantiationException) {
+              logger.error("Failed to load extension ${provider.type().simpleName}", e)
+              return@forEach
```
valeraz, line 102:
```suggestion
              return @forEach
```

valeraz (approved):

PR #380, opened 2023-06-06
Fix bootstrap not using jvm vendor property

ZacSweers:


arpita184:
Overall it looks good but can you please add more details around the purpose of this PR. Why is it needed or what should be the behavior after this change? Thanks!

arpita184 (approved):

PR #382, opened 2023-06-06
Improve unit test OOM handling

ZacSweers:
- Write a crash log on failure via CrashOnOutOfMemoryError
- Don't try to cat the log and hprof files
- Write crash log to same location

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/unittest/UnitTests.kt
+++ b/slack-plugin/src/main/kotlin/slack/unittest/UnitTests.kt
@@ -222,13 +222,13 @@ internal object UnitTests {
             else -> project.rootProject.projectDir.absolutePath
           }
         jvmArgs(
-          "-XX:+HeapDumpOnOutOfMemoryError",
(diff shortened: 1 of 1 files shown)


slack-plugin/src/main/kotlin/slack/unittest/UnitTests.kt
```diff
@@ -222,13 +222,13 @@ internal object UnitTests {
             else -> project.rootProject.projectDir.absolutePath
           }
         jvmArgs(
-          "-XX:+HeapDumpOnOutOfMemoryError",
+          "-XX:+HeapDumpOnOutOfMemoryError", // Produce a heap dump when an OOM occurs
+          "-XX:+CrashOnOutOfMemoryError", // Produce a crash report when an OOM occurs
```
arpita184, line 226: Should you also be adding- ExitOnOutOfMemoryError?

ZacSweers: Nope, you would pick only one of exit or crash

PR #398, opened 2023-06-13
Stub side panel files from studio 

arpita184:
Stub side panel files from studio at : https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:whats-new-assistant/src/com/android/tools/idea/whatsnew/assistant/


ZacSweers:
Continued in #400

PR #427, opened 2023-06-28
Bump code-review workflow

fxchen:
Over the last day, I’ve made a few changes to how we’re prompting and including context to the github action.
Also I removed the “post errors by default” - so errors will not show up as comments.
In this workflow we also filter out `slack-oss-bot` similar to how we do in `kaldb` (I think this will filter correctly). Example comment: https://github.com/slackhq/kaldb/pull/588#issuecomment-1612152887

This addresses some of your comments from the earlier prototype!

### DIFF
--- a/.github/workflows/code-review.yml
+++ b/.github/workflows/code-review.yml
@@ -1,6 +1,13 @@
 # This workflow triggers a code review using the OpenAI GPT-3.5 Turbo model with a 16k context window
+name: Code Review
 
-on: [pull_request]
+on:
+  pull_request: # Filter out draft pull requests
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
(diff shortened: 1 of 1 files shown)


.github/workflows/code-review.yml
```diff
@@ -1,16 +1,23 @@
...
+      - synchronize
+      - ready_for_review
 
 jobs:
   code-review:
     runs-on: ubuntu-latest
     permissions:
       contents: read
       pull-requests: write
+    if: ${{ github.actor != 'slack-oss-bot' }}
```
ZacSweers, line 18: I still think the bot should ideally be able to recognize if a given PR is automated or just a dependency update, but I'm ok with this for now

ZacSweers (approved):

PR #429, opened 2023-06-29
Simplify CI

ZacSweers:
We don't really ever set other versions and having a consistent name is required in order to mark those checks as `required` for merge queue

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,20 +18,7 @@ concurrency:
 
 jobs:
   build:
-    name: JDK ${{ matrix.ci_java_version }} / AGP ${{ matrix.ci_agp_version }} / Kotlin ${{ matrix.ci_kotlin_version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -18,32 +18,25 @@ concurrency:
 
 jobs:
   build:
-    name: JDK ${{ matrix.ci_java_version }} / AGP ${{ matrix.ci_agp_version }} / Kotlin ${{ matrix.ci_kotlin_version }}
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        ci_java_version: ['20']
```
arpita184, line 26: matrix.ci_java_version is still being used in 2 locations. Can you please resolve those references as well?

ZacSweers: Good catch 07f36eb

arpita184 (approved):

PR #444, opened 2023-07-05
Misc dependency rake improvements

ZacSweers:
See individual commits!

### DIFF
--- a/docs/dependency-rake.md
+++ b/docs/dependency-rake.md
@@ -35,3 +35,21 @@ that we currently do not do.
 ## Implementation
 
 The core implementation of DR lives in `DependencyRake.kt`.
+
+## Usage
+
+To run dependency rake in a project, use the below command
+
+```bash
+$ ./gradlew rakeDependencies -Pslack.gradle.config.enableAnalysisPlugin=true --no-configuration-cache
+```
+
+This will run all `rakeDependencies` tasks in the project. This task exists on all subprojects as well, but it 
+works best if all are run together.
+
+Sometimes dependency rake will try to replace identifiers with ones that are not present in any available
+version catalogs. Sometimes this is acceptable, but often times it can result in "missing" dependencies from 
+the build after it runs. To help fix these, DR will write all missing identifiers out to a build output file.
+
+For convenience, you can also run `./gradlew aggregateMissingIdentifiers -Pslack.gradle.config.enableAnalysisPlugin=true --no-configuration-cache`
+to run all dependency rake tasks and aggregate these missing identifiers into a root project build output file.
--- a/slack-plugin/src/main/kotlin/slack/dependencyrake/DependencyRake.kt
+++ b/slack-plugin/src/main/kotlin/slack/dependencyrake/DependencyRake.kt
@@ -72,8 +72,6 @@ constructor(objects: ObjectFactory, providers: ProviderFactory) : AbstractPostPr
   @get:InputFile
   abstract val buildFileProperty: RegularFileProperty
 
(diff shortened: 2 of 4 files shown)


docs/dependency-rake.md
```diff
@@ -35,3 +35,21 @@ that we currently do not do.
 ## Implementation
 
 The core implementation of DR lives in `DependencyRake.kt`.
+
+## Usage
+
+To run dependency rake in a project, use the below command
+
+```bash
+$ ./gradlew rakeDependencies -Pslack.gradle.config.enableAnalysisPlugin=true --no-configuration-cache
```
autonomousapps, line 44: `--no-configuration-cache` so sad, I should fix that.

docs/dependency-rake.md
```diff
@@ -35,3 +35,21 @@ that we currently do not do.
...
+```
+
+This will run all `rakeDependencies` tasks in the project. This task exists on all subprojects as well, but it 
+works best if all are run together.
+
+Sometimes dependency rake will try to replace identifiers with ones that are not present in any available
+version catalogs. Sometimes this is acceptable, but often times it can result in "missing" dependencies from 
+the build after it runs. To help fix these, DR will write all missing identifiers out to a build output file.
```
autonomousapps, line 52: I'm curious about this, can you elaborate?

slack-plugin/src/main/kotlin/slack/dependencyrake/DependencyRake.kt
```diff
@@ -129,7 +143,7 @@ constructor(objects: ObjectFactory, providers: ProviderFactory) : AbstractPostPr
         advices
           .filter { it.isRemove() }
           .filterNot { it.coordinates.identifier in MANAGED_DEPENDENCIES }
-          .associateBy { it.toDependencyString("UNUSED") }
+          .associateBy { it.toDependencyString("UNUSED", missingIdentifiers) }
```
autonomousapps, line 146: Also just curious about these magic strings.

slack-plugin/src/main/kotlin/slack/dependencyrake/DependencyRake.kt
```diff
@@ -110,16 +117,23 @@ constructor(objects: ObjectFactory, providers: ProviderFactory) : AbstractPostPr
     val redundantPlugins = projectAdvice.pluginAdvice
     val advices: Set<Advice> = projectAdvice.dependencyAdvice
     val buildFile = buildFileProperty.asFile.get()
+    val missingIdentifiers = mutableSetOf<String>()
```
autonomousapps, line 120: missing identifiers just means missing from your version catalog, right?

autonomousapps (approved):
Should I make a comment about the lack of tests? :D

docs/dependency-rake.md
ZacSweers: Basically we have a lot of transitive deps that DAGP offers as "add" advice but don't exist in our version catalogs, so this is intended to help make setting those up easier

slack-plugin/src/main/kotlin/slack/dependencyrake/DependencyRake.kt
ZacSweers: yup!

slack-plugin/src/main/kotlin/slack/dependencyrake/DependencyRake.kt
ZacSweers: Just debug logging!

PR #474, opened 2023-07-19
Don't show "What's New" Panel if no Changelog text is present

kateliu20:
Before this change, if there were no new entries a panel would still appear in the IDE:
Screenshot 2023-07-19 at 4 32 55 PM

Now if there is no text, the panel just simply won't show up (the xml file is open to show that a last date entry was stored) and there will be no tool window registered: 
Screenshot 2023-07-19 at 4 06 31 PM

I also realized that in another PR I may have accidentally merged a change with the settings:
Screenshot 2023-07-19 at 4 03 23 PM

So I reverted that to what it was originally (I am still looking into the wrap text so that will be in another PR):
Screenshot 2023-07-19 at 4 30 55 PM

### DIFF
--- a/skate-plugin/src/main/kotlin/com/slack/sgp/intellij/SkateService.kt
+++ b/skate-plugin/src/main/kotlin/com/slack/sgp/intellij/SkateService.kt
@@ -47,6 +47,7 @@ class SkateProjectServiceImpl(private val project: Project) : SkateProjectServic
     val changeLogString = VfsUtil.loadText(changeLogFile)
 
     // Don't show the tool window if the parsed changelog is blank
+    // Changelog is parsed
     val parsedChangelog = ChangelogParser.readFile(changeLogString, changelogJournal.lastReadDate)
     if (parsedChangelog.changeLogString.isNullOrBlank()) return
 
@@ -65,7 +66,7 @@ class SkateProjectServiceImpl(private val project: Project) : SkateProjectServic
       val parentDisposable = Disposer.newDisposable()
 
       WhatsNewPanelFactory()
-        .createToolWindowContent(toolWindow, project, changeLogString, parentDisposable)
+        .createToolWindowContent(toolWindow, project, parsedChangelog, parentDisposable)
 
       toolWindow.show()
     }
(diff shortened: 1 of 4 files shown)


skate-plugin/src/main/kotlin/com/slack/sgp/intellij/SkateService.kt
```diff
@@ -36,14 +36,21 @@ interface SkateProjectService {
...
+    val changelogJournal = project.service<ChangelogJournal>()
+
     if (!settings.isWhatsNewEnabled) return
     val projectDir = project.guessProjectDir() ?: return
+
     val changeLogFile = VfsUtil.findRelativeFile(projectDir, settings.whatsNewFilePath) ?: return
     val changeLogString = VfsUtil.loadText(changeLogFile)
+
+    // Don't show the tool window if the parsed changelog is blank
+    val parsedChangelog = ChangelogParser.readFile(changeLogString, changelogJournal.lastReadDate)
```
ZacSweers, line 50: Shouldn't we be only passing this in to the window below? Looks like we're still passing in the full changelogstring

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/ui/SkateConfigUI.kt
```diff
@@ -38,7 +38,7 @@ internal class SkateConfigUI(
 
   private fun Panel.checkBoxRow() {
     row(SkateBundle.message("skate.configuration.enableWhatsNew.title")) {
-      checkBox("skate.configuration.enableWhatsNew.description")
+      checkBox(SkateBundle.message("skate.configuration.enableWhatsNew.description"))
```
ZacSweers, line 41: 👍

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/SkateService.kt
kateliu20: My logic here is to check if the `parsedChangelog` would be empty, because if it is, it will return before the creation of the tool window. So I put it before

ZacSweers: Right but isn't the window always going to show the whole `changeLogString` each time currently, even if we only want to show a partial of that?

kateliu20: I think right now it just shows the `parsedChangelog` and I'm passing in the `changeLogString `to parse it, right? Do you mean that the change I'm making is always going to display the full string?

ZacSweers (approved):

PR #491, opened 2023-07-27
Cosmetic Bug Fix for Markdown File Rendering (with and without italics) 

kateliu20:
Account for both italicized and not italicized dates in the "What's New" files.

In the CHANGELOG file, the dates are wrapped in "_" and the WHATSNEW file, the dates are not.

Screenshot 2023-07-27 at 1 08 17 PM

Screenshot 2023-07-27 at 1 08 34 PM


arpita184 (approved):
Good catch

PR #503, opened 2023-08-07
Use one version code for all apks

linhpha:
We can move away from having separate version code for each build type now that we're publishing app bundle

This change should also help resolve issues with AGP 8.1 upgrade

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/gradle/ApkVersioningPlugin.kt
+++ b/slack-plugin/src/main/kotlin/slack/gradle/ApkVersioningPlugin.kt
@@ -129,7 +129,7 @@ internal class ApkVersioningPlugin : Plugin<Project> {
           val mappedVersionCodeProvider =
             versionCodeProvider.map { rawCode ->
               @Suppress("MagicNumber")
-              ApkVersioning.VERSION_CODES.getValue(variantOutput.abiString) * 10000000 + rawCode
+              ApkVersioning.VERSION_CODE * 10000000 + rawCode
             }
           variantOutput.versionCode.setDisallowChanges(mappedVersionCodeProvider)
         }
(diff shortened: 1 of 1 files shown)


slack-plugin/src/main/kotlin/slack/gradle/ApkVersioningPlugin.kt
```diff
@@ -129,7 +122,7 @@ internal class ApkVersioningPlugin : Plugin<Project> {
           val mappedVersionCodeProvider =
             versionCodeProvider.map { rawCode ->
               @Suppress("MagicNumber")
-              ApkVersioning.VERSION_CODES.getValue(variantOutput.abiString) * 10000000 + rawCode
+              ApkVersioning.VERSION_CODE * 10000000 + rawCode
```
ZacSweers, line 125: Let's drop this `10000000` and the `VERSION_CODE` magic constant and handle it internally in the slack repo's gradle.properties, no reason to impose this heh. Then we can just pass `versionCodeProvider` directly to the `variantOutput.versionCode`

linhpha: Removed the calculation here https://github.com/slackhq/slack-gradle-plugin/pull/503/commits/37159ee0888a58413fae0c74dae5eada22871532

Per this change, we'll need to update the version code in [release.version](https://github.com/tinyspeck/slack-android-ng/blob/main/ci/release.version) to 90012557 right?

ZacSweers (approved):
Lgtm. One possibly tricky thing is that we either need to do a release with this cherry-picked onto a branch from the previous non-kotlin 1.9 update or do them all at once. What do you think?

slack-plugin/src/main/kotlin/slack/gradle/ApkVersioningPlugin.kt
ZacSweers: Yup exactly

linhpha:
> Lgtm. One possibly tricky thing is that we either need to do a release with this cherry-picked onto a branch from the previous non-kotlin 1.9 update or do them all at once. What do you think?

I think I'd prefer the latter cause it's easier 😅 I've been testing locally the upgrade with this change and it should work 🤞

ZacSweers:
Kk I'll take care of it

ZacSweers (approved):

PR #508, opened 2023-08-08
Create a skeleton for Linker Plugin

arpita184:
This is a skeleton for a linker plugin - A plugin that will link feature flags to their server counterparts.


ZacSweers (commented):
We want to add this functionality to the skate plugin, not as a separate plugin :)

PR #500, opened 2023-07-31
Add PluginRepositoryAuthProvider for Custom Repository Authentication 

kateliu20:
Solution from: https://youtrack.jetbrains.com/issue/IDEA-315487/PluginRepositoryAuthProvider-is-broken-in-2022.3-making-custom-repository-authentication-impossible#focus=Comments-27-6990965.0-0

PR #509, opened 2023-08-08
Fix authenticator plugin

ZacSweers:
Continuing from #500. Also had to remove the okio dependency as it appears to conflict with dependencies coming from the intellij plugin now, so using standard java to do it. Tested this with studio giraffe (the bugged version) and it works now


arpita184 (approved):
Looks good. Can you please add a documentation around why this hotfix was needed for future us. Thanks!

PR #529, opened 2023-08-15
Improve Skate changelog parsing logic

ZacSweers:
This PR rewrites changelog parser logic to use a sequence instead

This makes the changelog parser logic a bit more robust and lazy, while also fixing some specific issues around parsing dates (previously it assumed dates were italicized).

### DIFF
--- a/skate-plugin/src/test/kotlin/com/slack/sgp/intellij/ChangeLogParserTest.kt
+++ b/skate-plugin/src/test/kotlin/com/slack/sgp/intellij/ChangeLogParserTest.kt
@@ -21,9 +21,9 @@ import org.junit.Test
 
 class ChangeLogParserTest {
   @Test
-  fun `no entries and null changelogstring`() {
+  fun `no entries and null changelog string`() {
     val (changeLogString, latestEntry) = ChangelogParser.readFile("", null)
-    assertThat(changeLogString).isEmpty()
+    assertThat(changeLogString).isNull()
     assertThat(latestEntry).isEqualTo(LocalDate.now())
   }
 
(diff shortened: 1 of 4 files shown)


skate-plugin/src/test/kotlin/com/slack/sgp/intellij/ChangeLogParserTest.kt
```diff
@@ -21,9 +21,9 @@ import org.junit.Test
 
 class ChangeLogParserTest {
```
ZacSweers, line 22: An issue with the previous tests is that they tested the format of the `CHANGELOG.md` file in this repo, which used this format

```
0.1.0
-----

_2023-05-05_

- Some update
```

when the actual format we parse in `WHATSNEW.md` files is

```

```
2023-05-05
----------

- Some update
```
```

jhperezd: Hey @ZacSweers , I do see Italics `_` in WHATSNEW.md in the Android repo. Should that be changed? There is only one entry at the moment:
------------------------
What's New
==========

_2023-08-08_
------------

### Android Studio ....
------------------------

ZacSweers: Yep we'll need to fix those, they were added because of this confusion I think heh

jhperezd: 😅 kk, I'll approve 👍

jhperezd (approved):
👍

PR #533, opened 2023-08-16
Add common IJ plugins we use as dependencies

### LINKED ISSUE #510: Add standard IDE plugins as dependencies in skate
Things we use like ktfmt, google-java-format, kotlin, detekt, android, etc would be good to have. Bonus if we can auto-configure them too (i.e. like enabling the formatters, setting their styles, etc).

ZacSweers:
Resolves #510


linhpha (approved):

ZacSweers:
Closing this out for now as this didn't work the way I expected

PR #547, opened 2023-08-25
Add buildToolsVersionOverride

ZacSweers:
Sometimes we need to explicitly define a newer version for these to pick up fixes like https://issuetracker.google.com/issues/237299698


slack-plugin/src/main/kotlin/slack/gradle/StandardProjectConfigurations.kt
```diff
@@ -830,7 +831,7 @@ internal class StandardProjectConfigurations(
                       // Skip dashes and underscores. We could camelcase but it looks weird in a
                       // package name
                       '-',
-                      '_' -> null
+                      '_', -> null
```
arpita184, line 834: Please verify this comma "," is correctly placed. This syntax looks off to me.

arpita184 (approved):
1 small nit.

PR #576, opened 2023-09-20
Feature Flag Annotator 

arpita184:
What is this PR for?

This PR enhances the user experience by implementing automated linking of feature flags. It enables users to quickly access feature flag links through inspections for corresponding enum entries.

Key Features:
Automated Scanning: Detects enums annotated with the fully-qualified name (com.example.FeatureFlag) in *Feature.kt files.
Direct Linking: Adds inspections to the identified enum entries for seamless navigation to the feature flag links.

Screenshot 2023-09-28 at 12 43 49 PM

How to enable this plugin?
- Navigate to the plugin's settings screen.
- Activate the 'feature flag linking' option.
- Specify the fully qualified name for enum identification.
- Input the base URL, which will be prefixed to the flag key for forming the complete link.
Screenshot 2023-09-28 at 1 03 51 PM


skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
```diff
@@ -0,0 +1,100 @@
...
+  private fun findAnnotatedEnums(psiFile: Any): List<String> {
+    val annotatedEnumEntries = mutableListOf<String>()
+
+    fun addIfAnnotatedEnum(element: Any) {
+      if (isKtEnumEntry(element) && hasFeatureFlagAnnotation(element)) {
+        element.javaClass
+          .getMethod("getName")
+          .invoke(element)
+          .takeIf { it is String }
+          ?.let { annotatedEnumEntries.add(it as String) }
```
ZacSweers, line 63: This isn't quite correct. The key logic will use the key defined in the `@FeatureFlag` annotation if it's not blank, and otherwise fallback to the lowercase (US locale) entry name as a default.

skate-plugin/build.gradle.kts
```diff
@@ -17,7 +17,10 @@ repositories { mavenCentral() }
 
 // Configure Gradle IntelliJ Plugin
 // Read more: https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html
-intellij { plugins.add("org.intellij.plugins.markdown") }
+intellij {
+  plugins.add("org.intellij.plugins.markdown")
+  plugins.add("org.jetbrains.kotlin")
```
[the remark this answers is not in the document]
arpita184: This is to address `NoClassDefFoundError` error. This error appears when I run tests or plugin
  The Skate (id=com.slack.intellij.skate, path=~/slack-gradle-plugin/skate-plugin/build/idea-sandbox/plugins-test/skate-plugin, version=0.1.0) plugin Plugin 'Skate' requires plugin 'org.jetbrains.kotlin' to be installed


java.lang.NoClassDefFoundError: org/jetbrains/kotlin/idea/KotlinLanguage

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/featureFlagAnnotator.kt
```diff
@@ -0,0 +1,158 @@
...
+    if (flags.isNullOrEmpty()) {
+      return emptyList()
+    }
+    val baseUrl = psiFile.project.service<SkatePluginSettings>().featureFlagBaseUrl.orEmpty()
+
+    return flags.mapNotNull { flag ->
+      val textRange =
+        ApplicationManager.getApplication().runReadAction<TextRange?> {
+          findTextRangeForFlag(psiFile, flag)
+        }
```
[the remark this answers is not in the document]
arpita184: I was getting this error - java.lang.Throwable: Read access is allowed from inside read-action (or EDT) only (see com.intellij.openapi.application.Application.runReadAction())

Some code is trying to read from the PSI without being inside an appropriate read-action so I needed to wrap a reading operation like this to address the error

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/featureFlagAnnotator.kt
```diff
@@ -0,0 +1,158 @@
...
+  override fun invoke(project: Project, editor: Editor?, file: PsiFile?) {
+    BrowserUtil.browse(URI(url))
+  }
+
+  override fun startInWriteAction(): Boolean {
+    return false
+  }
+}
+
+class MyGutterIconRenderer(private val url: String) : GutterIconRenderer() {
```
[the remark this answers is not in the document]
arpita184: Curious, is there any harm in having both? I like gutter icon as a direct visual indicator as intention can be missed if one didn't know about it.

skate-plugin/build.gradle.kts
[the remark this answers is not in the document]
ZacSweers: The kotlin plugin is bundled anyway though, so I'm curious what difference this makes?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/featureFlagAnnotator.kt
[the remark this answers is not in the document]
ZacSweers: Gotcha. Is it only possible to do a read action from the application instance? We saw similar with what's new but iirc that uses a more contextual mechanism

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/featureFlagAnnotator.kt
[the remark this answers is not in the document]
ZacSweers: Yes, it's redundant to have both and we should be very conservative about adding more information to an already information-dense UI

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
```diff
@@ -0,0 +1,100 @@
...
+ * Responsible for extracting feature flags. Searches for enum entries annotated with 'FeatureFlag'
+ * to identify feature flags.
+ */
+class FeatureFlagExtractor {
+
+  private val log: Logger = Logger.getInstance(FeatureFlagExtractor::class.java)
+
+  // Caches the feature flags for a given PSI file to optimize repeated lookups
+  companion object {
+    private val featureFlagCache = mutableMapOf<com.intellij.psi.PsiFile, List<String>>()
```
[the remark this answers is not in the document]
arpita184: No reason except a miss on my part..Fixed with [38de1b7](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/38de1b709b7bac4cde025d7a37902f7d349c9ff4)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/featureFlagAnnotator.kt
```diff
@@ -0,0 +1,158 @@
...
+import com.intellij.util.IconUtil
+import com.slack.sgp.intellij.SkatePluginSettings
+import com.slack.sgp.intellij.TEST_KOTLIN_LANGUAGE_ID_KEY
+import com.slack.sgp.intellij.util.isLinkifiedFeatureFlagsEnabled
+import java.net.URI
+import javax.swing.Icon
+import org.jetbrains.kotlin.idea.KotlinLanguage
+
+class FeatureFlagAnnotator : ExternalAnnotator<PsiFile, List<FeatureFlagSymbol>>() {
+  private val extractor = FeatureFlagExtractor()
```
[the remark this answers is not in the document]
arpita184: No we don't. Updated: [7843719](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/784371932e076fc1fe70b972b1dc386e7119fea0)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/featureFlagAnnotator.kt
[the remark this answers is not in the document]
arpita184: Updated: [7843719](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/784371932e076fc1fe70b972b1dc386e7119fea0)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/featureFlagAnnotator.kt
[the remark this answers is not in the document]
arpita184: Removed [7843719](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/784371932e076fc1fe70b972b1dc386e7119fea0)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/featureFlagAnnotator.kt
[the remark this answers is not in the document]
ZacSweers: In the future let's split those into separate commits so it's easier to map changes to comments they address 👍

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotator.kt
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  override fun doAnnotate(collectedInfo: PsiFile): List<FeatureFlagSymbol> {
+    if (!collectedInfo.project.isLinkifiedFeatureFlagsEnabled() || !isKotlinFile(collectedInfo)) {
+      return emptyList()
+    }
+
+    val flagsForFile = extractor.getFeatureFlagsForPsiFile(collectedInfo)
+    return transformToFeatureFlagSymbols(collectedInfo, flagsForFile)
+  }
```
ZacSweers, line 58: I'm curious why both of these functions need to be implemented?

arpita184: As per the documentation 

```
Implemented by a custom language plugin to process the files in a language by an external annotation tool ("linter"). External annotators are expected to be (relatively) slow and are started after regular annotators have completed their work.
Annotators work in three steps:
collectInformation(PsiFile, Editor, boolean) is called to collect some data about a file needed for launching a tool
collected data is passed to doAnnotate which executes a tool and collects highlighting data
highlighting data is applied to a file by apply
```

Based on that my understanding it splits the responsibilities in a clear way. In our case if collectInformation thinks it is not a featureFile then we won't even need to call doAnnotate

ZacSweers: hmmm, I went and read the [source](https://github.com/JetBrains/intellij-community/blob/master/platform/analysis-api/src/com/intellij/lang/annotation/ExternalAnnotator.java) and you omitted some details. Namely that `collectInformation` is called within a read action and `doAnnotate` is not and should be for long-running tasks.

After reading it there, I believe what you should be doing here is this
- Change the first generic to `List<FeatureFlagSymbol>`. This is computed in `collectInformation` and, when returned in that function, will be passed on to `doAnnotate()`. Currently you use `PsiFile` and just read it once and use it as a cache key in `extractor` in `doAnnotate()`.
- It seems like `!collectedInfo.project.isLinkifiedFeatureFlagsEnabled() || !isKotlinFile(collectedInfo)` should be moved to the `collectInformation()` function.
- `doAnnotate` doesn't need to do any extra processing as far as I can tell, so seems it can return the collected data and leave it on for the `apply()` function to use

With that, you can also remove the cache from `FeatureFlagExtractor` and just make it an `object` class rather than instantiable as it's purely functional in nature.

skate-plugin/src/test/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotatorTest.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.intellij.psi.PsiFile
+import com.intellij.testFramework.fixtures.LightPlatformCodeInsightFixture4TestCase
+import com.slack.sgp.intellij.SkatePluginSettings
+import org.junit.Test
+
+class FeatureFlagAnnotatorTest : LightPlatformCodeInsightFixture4TestCase() {
+  private val fileContent =
+    """
+            @FeatureFlags(ForComplianceFeature::class)
+            enum class ComplianceFeature(override val dependencies: Set<FeatureFlagEnum> = emptySet()) :
```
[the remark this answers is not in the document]
arpita184: [a945d8c](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/a945d8c1683f019ba3b86066d5997a10e122b8d9)

skate-plugin/src/test/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotatorTest.kt
```diff
@@ -0,0 +1,74 @@
...
+
+import com.google.common.truth.Truth.assertThat
+import com.intellij.openapi.components.service
+import com.intellij.psi.PsiFile
+import com.intellij.testFramework.fixtures.LightPlatformCodeInsightFixture4TestCase
+import com.slack.sgp.intellij.SkatePluginSettings
+import org.junit.Test
+
+class FeatureFlagAnnotatorTest : LightPlatformCodeInsightFixture4TestCase() {
+  private val fileContent =
```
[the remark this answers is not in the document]
arpita184: [533826c](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/533826c9d66876ebec61e66beeb910c2d452aeb7)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/SkatePluginSettings.kt
```diff
@@ -38,8 +38,22 @@ class SkatePluginSettings : SimplePersistentStateComponent<SkatePluginSettings.S
...
     }
 
+  var isLinkifiedFeatureFlagsEnabled: Boolean
+    get() = state.isLinkifiedFeatureFlagsEnabled
+    set(value) {
+      state.isLinkifiedFeatureFlagsEnabled = value
+    }
+
+  var featureFlagBaseUrl: String?
+    get() = state.featureFlagBaseUrl ?: "https://houston.tinyspeck.com/experiments/all"
```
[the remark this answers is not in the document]
arpita184: [69f3cf4](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/69f3cf42ecf827e82d4fcee2f3a86b1a2cd6a329)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotator.kt
arpita184: [252b25b](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/252b25b00fef199674f7b82f9895f3f943dfa6fc)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
arpita184: No longer relevant

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/PsiElementExtensions.kt
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+fun PsiElement.getChildren(elementType: String): List<PsiElement> {
+  return this.childrenOfType<PsiElement>().filter { it.node.elementType.toString() == elementType }
+}
+
+fun PsiElement.getAnnotationEntries(): List<PsiElement> {
+  return this.getFirstChild("MODIFIER_LIST")?.getChildren("ANNOTATION_ENTRY") ?: listOf()
+}
```
ZacSweers, line 33: This should only be available on class types, not any element. It should use strongly-typed APIs to get class members that are enum entry types. Here's an example: https://github.com/slackhq/slack-lints/blob/main/slack-lint-checks/src/main/java/slack/lint/MoshiUsageDetector.kt#L1019

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotator.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  private fun isKotlinFile(psiFile: PsiFile): Boolean =
+    psiFile.language.id == KotlinLanguage.INSTANCE.id ||
+      psiFile.getUserData(TEST_KOTLIN_LANGUAGE_ID_KEY) == KotlinLanguage.INSTANCE.id
+
+  private fun isKotlinFeatureFile(psiFile: PsiFile): Boolean = psiFile.name.endsWith("Feature.kt")
+}
+
+class FeatureFlagSymbol(val element: PsiElement, val url: String)
```
ZacSweers, line 82: This should use a stronger type than `PsiElement`

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
ZacSweers: I think it is, the updated code still does not account for a `key` value

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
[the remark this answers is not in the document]
ZacSweers: The new impl does not use reflection, but it does use `PsiElement` everywhere, which is not much better as it's resulting in stringy comparison checks. Please try to use strongly-typed APIs like `PsiClass`, `UClass`, etc. The only strings you should ever need to check are names, not PSI metatada

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/PsiElementExtensions.kt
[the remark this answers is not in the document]
arpita184: I am currently working on that, almost there. Wanted to address reflection removal in 2 parts -  1) Remove all the reflection and make sure current approach works 2) make it typeSafe

skate-plugin/build.gradle.kts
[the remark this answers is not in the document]
arpita184: I need to still dig further into this. Tests do not seem to find bundled Kotlin.

But plugin does have Kotlin bundled as I verified by installing a plugin on Android studio. If we actually include - `plugins.add("org.jetbrains.kotlin")` in the build.gradle.kts then it will result into a classLoader issue at a runtime for a plugin so this needs to be sorted out before merging the PR.

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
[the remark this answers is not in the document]
arpita184: [3d79394](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/3d793946ee4cb54b388a873d84ce7576699e5376)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/PsiElementExtensions.kt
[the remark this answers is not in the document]
arpita184: [3d79394](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/3d793946ee4cb54b388a873d84ce7576699e5376)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
arpita184: Being handled [here](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/3d793946ee4cb54b388a873d84ce7576699e5376#diff-0073d4ea07ce931d766fbad1989d075c4289e50411408234e22459bde49a29bdR45)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/PsiElementExtensions.kt
```diff
@@ -16,34 +16,31 @@
...
 }
 
-fun PsiElement.getEnumIdentifier(): PsiElement? {
-  return this.getFirstChild("IDENTIFIER")
+fun KtAnnotationEntry.getKeyArgumentValue(key: String): String? {
+  val argumentExpression =
+    this.valueArguments
+      .firstOrNull { it.getArgumentName()?.asName?.asString() == key }
+      ?.getArgumentExpression()
+  return (argumentExpression as? PsiElement)?.toUElement(UExpression::class.java)?.evaluate()
```
ZacSweers, line 40: hmmm you shouldn't need to do this first `as? PsiElement`, can you find another way?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/PsiElementExtensions.kt
```diff
@@ -16,34 +16,31 @@
...
-fun PsiElement.isEnum(): Boolean {
-  return this.node?.elementType?.toString() == "ENUM_ENTRY"
+fun KtEnumEntry.getAnnotation(name: String): KtAnnotationEntry? {
+  return this.getAnnotationEntries().firstOrNull { it.shortName?.asString() == name }
 }
 
-fun PsiElement.hasAnnotation(name: String): Boolean {
-  return this.getAnnotationEntries().any { it.getAnnotationName() == name }
+fun PsiElement.getEnumIdentifier(): LeafPsiElement? {
+  return this.childrenOfType<LeafPsiElement>().firstOrNull { it.elementType == KtTokens.IDENTIFIER }
```
ZacSweers, line 32: Is this necessary? I think `KtEnumEntry` just extents `KtClass` and you should be able to get the name like any other class.

arpita184: Getting name is easy as you pointed out but then for annotation there is range() which takes textRange, ASTNode or PsiElement. If I get a name then I will I will have to read a file to find a textRange as follow and pass as .range(symbol.textRange)

```
  private fun findTextRangeForFlag(psiFile: PsiFile, flag: String): TextRange? {
    val elements = PsiTreeUtil.findChildrenOfType(psiFile, PsiElement::class.java)
    return elements.firstOrNull { it.text == flag }?.textRange
  } 
```
  So I wanted to keep it simply and directly pass LeadPsiElement which is of type PsiElement as 
  .range(symbol.element as PsiElement)
  
  Thoughts?

ZacSweers: that's not keeping it simple as it's doing manual AST parsing. As I mentioned in a few other spots, please take a look at similar processing of enums we do in the slack-lints repo. Here's a good example https://github.com/slackhq/slack-lints/blob/059b7f88083089e3ab959e454fe296e435c5c53b/slack-lint-checks/src/main/java/slack/lint/MoshiUsageDetector.kt#L1018

Once you have the `UEnumConstant`, it has a simple `getName()` function on it that returns a string.

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/PsiElementExtensions.kt
arpita184: I did `as? PsiElement` to make compiler happy as without it I get following error. So far the only way I found to evaluate() expression was to convert the element to UExpression.

```
e: file:///Users/arpita.patel/slack-gradle-plugin/skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/PsiElementExtensions.kt:40:30 Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
public fun PsiElement?.toUElement(): UElement? defined in org.jetbrains.uast
```

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotator.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.jetbrains.kotlin.psi.KtFile
+
+class FeatureFlagAnnotator : ExternalAnnotator<List<FeatureFlagSymbol>, List<FeatureFlagSymbol>>() {
+
+  override fun collectInformation(file: PsiFile): List<FeatureFlagSymbol> {
+    if (
+      !file.project.isLinkifiedFeatureFlagsEnabled() ||
+        file !is KtFile ||
+        !isKotlinFeatureFile(file)
+    ) {
```
ZacSweers, line 37: Can you pull this out to a variable first for readability? Nesting inside an if expression gets hard to read

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
```diff
@@ -0,0 +1,74 @@
...
+   */
+  fun extractFeatureFlags(psiFile: PsiFile): List<FeatureFlagSymbol> {
+    // Ensure baseUrl and flagAnnotation are not empty when the feature flag linkifying is enabled
+    // Ensure baseUrl ends with query param - "?q="
+    val settings = psiFile.project.service<SkatePluginSettings>()
+    val baseUrl = settings.featureFlagBaseUrl.orEmpty()
+    val flagAnnotation = settings.featureFlagAnnotation.orEmpty()
+    require(baseUrl.isNotBlank()) { BASE_URL_EMPTY_ERROR }
+    require(baseUrl.endsWith("?q=")) { BASE_URL_QUERY_PARAM_ERROR }
+    require(flagAnnotation.isNotBlank()) { ANNOTATION_EMPTY_ERROR }
```
ZacSweers, line 52: Do these checks sooner, this is too late and we should really do it in the same context where they enable this feature

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
```diff
@@ -0,0 +1,74 @@
...
+   * @param psiFile The PSI representation of the file to process.
+   */
+  fun extractFeatureFlags(psiFile: PsiFile): List<FeatureFlagSymbol> {
+    // Ensure baseUrl and flagAnnotation are not empty when the feature flag linkifying is enabled
+    // Ensure baseUrl ends with query param - "?q="
+    val settings = psiFile.project.service<SkatePluginSettings>()
+    val baseUrl = settings.featureFlagBaseUrl.orEmpty()
+    val flagAnnotation = settings.featureFlagAnnotation.orEmpty()
+    require(baseUrl.isNotBlank()) { BASE_URL_EMPTY_ERROR }
+    require(baseUrl.endsWith("?q=")) { BASE_URL_QUERY_PARAM_ERROR }
```
ZacSweers, line 51: This isn't something we should enforce, we should just promise to prefix the base URL to the resolved key

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
```diff
@@ -0,0 +1,74 @@
...
+      .allClassesAndInnerClasses()
+      .filter { it.isEnum }
+      .flatMap { enumClass -> enumClass.uastDeclarations.filterIsInstance<UEnumConstant>() }
+      .mapNotNull { enumConstant ->
+        enumConstant.findAnnotation(flagAnnotation)?.let { flagAnnotation ->
+          val key =
+            flagAnnotation.findAttributeValue("key")?.evaluateString()?.takeUnless {
+              it.trim().isBlank()
+            } ?: enumConstant.name.lowercase(Locale.US)
+          val textRange = enumConstant.sourcePsi?.textRange ?: return@mapNotNull null
```
ZacSweers, line 66: nit: let's switch the order here so we return early if we can't resolve the range

```suggestion
          val textRange = enumConstant.sourcePsi?.textRange ?: return@mapNotNull null
          val key =
            flagAnnotation.findAttributeValue("key")?.evaluateString()?.takeUnless {
              it.trim().isBlank()
            } ?: enumConstant.name.lowercase(Locale.US)
```

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/SkatePluginSettings.kt
```diff
@@ -55,10 +55,31 @@ class SkatePluginSettings : SimplePersistentStateComponent<SkatePluginSettings.S
       state.translatorFileNameSuffix = value
     }
 
+  var isLinkifiedFeatureFlagsEnabled: Boolean
```
ZacSweers, line 58: Can you add these to the settings UI and do error handling for missing urls/annotations there?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
arpita184: [dacb0bc](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/dacb0bc36c03ecd9a4fa4f1e49f4cba8bec0f91d)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotator.kt
arpita184: [d93844e](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/d93844e0a024067cd817e125f9a3144a414d2618)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
arpita184: [d1a586e](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/d1a586ee7e7422421220df56cd4b6c5709b1eed5)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
arpita184: ✅

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagExtractor.kt
```diff
@@ -0,0 +1,100 @@
...
+      element.javaClass.methods
+        .find { it.name == "getChildren" }
+        ?.let { method -> (method.invoke(element) as? Array<*>)?.forEach { recurse(it!!) } }
+    }
+    recurse(psiFile)
+    return annotatedEnumEntries
+  }
+
+  /**
+   * Determines if a given PSI element is an enum entry annotated with 'FeatureFlag'.
```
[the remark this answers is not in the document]
arpita184: [d1a586e](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/d1a586ee7e7422421220df56cd4b6c5709b1eed5)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/PsiElementExtensions.kt
arpita184: [ec81f44](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/ec81f4417cd0409bed9ab368d3bdb606b12ea717)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotator.kt
arpita184: [ec81f44](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/ec81f4417cd0409bed9ab368d3bdb606b12ea717#diff-0073d4ea07ce931d766fbad1989d075c4289e50411408234e22459bde49a29bdR76)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/PsiElementExtensions.kt
arpita184: [ec81f44](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/ec81f4417cd0409bed9ab368d3bdb606b12ea717)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/PsiElementExtensions.kt
arpita184: [ec81f44](https://github.com/slackhq/slack-gradle-plugin/pull/576/commits/ec81f4417cd0409bed9ab368d3bdb606b12ea717)
Thanks for suggestions and hint that I should just covert the PsiFile directly to UFile early and start using UAST APIs directly from the start instead of `as? PsiElement` and other lower level API calls first

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/ui/SkateConfigUI.kt
```diff
@@ -71,4 +76,53 @@ internal class SkateConfigUI(
         .enabled(settings.isWhatsNewEnabled)
     }
   }
+
+  private fun Panel.featureFlagSettings() {
```
ZacSweers, line 80: Nice 👍

skate-plugin/src/main/resources/messages/skateBundle.properties
```diff
@@ -4,3 +4,8 @@ skate.configuration.choosePath.dialog.title=Choose Changelog File
 skate.configuration.enableWhatsNew.title=Enable What's New Panel
 skate.configuration.enableWhatsNew.description=If enabled, shows the What's New Panel if there are new changes detected in a specified changelog file.
 skate.modelTranslator.description=Generate translator body
+skate.configuration.enableFeatureFlagLinking.title=Enable Feature Flag Linking
+skate.configuration.enableFeatureFlagLinking.description=If enabled, feature flag URL will be shown in the quick documentation tooltip.
```
ZacSweers, line 8: This isn't quite right

```suggestion
skate.configuration.enableFeatureFlagLinking.description=If enabled, show inspections for opening feature flags.
```

skate-plugin/src/main/resources/messages/skateBundle.properties
```diff
@@ -4,3 +4,8 @@ skate.configuration.choosePath.dialog.title=Choose Changelog File
...
+skate.configuration.featureFlagBaseUrl.title=Enter URL to be added as a prefix to flag annotation
```
ZacSweers, line 9:
```suggestion
skate.configuration.featureFlagBaseUrl.title=Base URL for feature flag linking. The flag key will be appended to the end of this.
```

skate-plugin/src/main/resources/messages/skateBundle.properties
```diff
@@ -4,3 +4,8 @@ skate.configuration.choosePath.dialog.title=Choose Changelog File
...
+skate.configuration.featureFlagAnnotation.title=Feature flag annotation name
```
ZacSweers, line 10:
```suggestion
skate.configuration.featureFlagAnnotation.title=Feature flag annotation's fully-qualified name (com.example.FeatureFlag)
```

skate-plugin/src/main/resources/messages/skateBundle.properties
```diff
@@ -4,3 +4,8 @@ skate.configuration.choosePath.dialog.title=Choose Changelog File
...
+skate.configuration.featureFlagFieldEmpty.error=Field cannot be empty when feature flag Linking is enabled
```
ZacSweers, line 11:
```suggestion
skate.configuration.featureFlagFieldEmpty.error=Field cannot be empty when feature flag linking is enabled
```

ZacSweers (commented):
Almost there! Few more nits on the settings wording

ZacSweers (approved):
💥

PR #582, opened 2023-09-22
Add support for AGP 8.3.0-alpha05+

ZacSweers:
Been a minute but this release introduced a breaking change that requires us to add a new AgpHandler type for SGP to work across multiple versions.


linhpha (approved):

linhpha (changes requested):
CI's failing with deprecated references to `AgpHandlerFactory` left in [slack/gradle/AgpHandlers.kt](https://github.com/slackhq/slack-gradle-plugin/blob/main/slack-plugin/src/main/kotlin/slack/gradle/AgpHandlers.kt). Might need to change it with the new `AgpHandler.Factory`

linhpha (approved):

PR #590, opened 2023-09-27
Add ProjectGen option to Android Studio Tools Menu

linhpha:
This PR adds the new `Slack Project Generator` option to Tools menu under Skate plugin. Upon clicking it, devs should see a new terminal pop up running the `./slackw` command to open the ProjectGen Desktop App

Demo:

https://github.com/slackhq/slack-gradle-plugin/assets/12748234/0b54364b-1f01-429d-8ac4-d39fd5b4943b


ZacSweers (commented):
Awesome! Do you think we can add a simple smoke test for this? Not sure how feasible it is since it pops open the shell widget to run.

Also are you planning to eventually port the whole implementation to the plugin itself so we don't need to shell out to a command?

linhpha:
> Awesome! Do you think we can add a simple smoke test for this? Not sure how feasible it is since it pops open the shell widget to run.


Yeah I couldn't think of a good way to test this. I could addd some tests to verify that the command passed into the shell widget is as expected, but the main work here is making sure the shell is doing the right job.

> 
> Also are you planning to eventually port the whole implementation to the plugin itself so we don't need to shell out to a command?

Was gonna ask you about that too :) Are we cool with moving project gen over here? It'd be neat to keep everything in one place

ZacSweers:
Yep let's eventually move it here, should fit right in since a lot of this is Java Swing

skate-plugin/build.gradle.kts
```diff
@@ -1,8 +1,9 @@
+
 import com.jetbrains.plugin.structure.base.utils.exists
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.nio.file.Paths
-import java.util.Locale
+import java.util.*
```
ZacSweers, line 5: you can fix your IDE settings to avoid this

Settings > Editor > Code Style > Kotlin > Imports

Set everything to 99 and remove package imports

skate-plugin/build.gradle.kts
```diff
@@ -92,4 +93,5 @@ dependencies {
   implementation(libs.bugsnag) { exclude(group = "org.slf4j") }
   testImplementation(libs.junit)
   testImplementation(libs.truth)
+  testImplementation("org.mockito:mockito-core:5.5.0")
```
ZacSweers, line 96: We don't use mockito in this project :). Let's stick to fakes and real implementations

skate-plugin/src/test/kotlin/com/slack/sgp/intellij/ProjectGenMenuActionTest.kt
```diff
@@ -0,0 +1,50 @@
...
+  override fun setUp() {
+    super.setUp()
+
+    // Reset relevant settings.
+    val settings = skatePluginSettings()
+    settings.isProjectGenMenuActionEnabled = true
+    settings.projectGenRunCommand = "echo Hello World"
+  }
+
+  private fun skatePluginSettings() = project.service<SkatePluginSettings>()
```
ZacSweers, line 36: I believe we have a helper function for this already in the project now you could use

skate-plugin/src/test/kotlin/com/slack/sgp/intellij/ProjectGenMenuActionTest.kt
```diff
@@ -0,0 +1,50 @@
...
+
+  fun testCommandLineParam() {
+    val action = ProjectGenMenuAction()
+    val mockTerminal = Mockito.mock(TerminalView::class.java)
+    val mockShellWidget = Mockito.mock(ShellTerminalWidget::class.java)
```
ZacSweers, line 41: If you're unable to fake these with real implementations, let's create a simple wrapper interface that can be easily faked in a test instead.

linhpha: Tried adding fakes here https://github.com/slackhq/slack-gradle-plugin/pull/590/commits/23e43eade27da39365dd38fe6a0af1cfba3ae5c9. Lmk what you think!

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/projectgen/TerminalViewWrapper.kt
```diff
@@ -0,0 +1,35 @@
...
+ */
+package com.slack.sgp.intellij.projectgen
+
+import java.io.IOException
+import org.jetbrains.plugins.terminal.TerminalView
+
+/* Wrapper around Jetbrains TerminalView to help simplify testing */
+class TerminalViewWrapper(
+  private val view: TerminalView,
+) : TerminalViewInterface {
```
ZacSweers, line 24: nit: I like calling things like this `Real<thing>`, i.e. `RealTerminalViewWrapper`. Then call the interface `TerminalViewWrapper` instead of the interface

skate-plugin/src/test/kotlin/com/slack/sgp/intellij/fakes/FakeTerminalViewWrapper.kt
```diff
@@ -0,0 +1,30 @@
...
+class FakeTerminalViewWrapper : TerminalViewInterface {
+  var commandRan: String? = null
+  var projectPath: String? = null
+  var tabName: String? = null
+
+  override fun executeCommand(command: String, projectPath: String?, tabName: String) {
+    this.commandRan = command
+    this.projectPath = projectPath
+    this.tabName = tabName
+  }
```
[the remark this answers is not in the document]
linhpha: Thanks for the suggestion! Updated!

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/projectgen/TerminalViewWrapper.kt
```diff
@@ -0,0 +1,35 @@
...
+  override fun executeCommand(command: TerminalCommand) {
+    try {
+      // Create new terminal window to run given command
+      view
+        .createLocalShellWidget(command.projectPath, command.tabName)
+        .executeCommand(command.command)
+    } catch (err: IOException) {
+      err.printStackTrace()
```
ZacSweers, line 32: Let's just throw the error here, that's a major bug and we don't want to quietly fail in this case

skate-plugin/src/test/kotlin/com/slack/sgp/intellij/ProjectGenMenuActionTest.kt
```diff
@@ -0,0 +1,58 @@
...
+  fun testTerminalViewNotRunningWhenActionDisabled() {
+    val action = ProjectGenMenuAction()
+    action.terminalViewWrapper = FakeTerminalViewWrapper()
+    project.service<SkatePluginSettings>().isProjectGenMenuActionEnabled = false
+
+    // Perform action
+    myFixture.testAction(action)
+
+    // Verify action didn't run any terminal command
+    (action.terminalViewWrapper as FakeTerminalViewWrapper).assertEmptyCommand()
```
ZacSweers, line 56: If you pull the fake out to a variable you don't need to cast it

```suggestion
    val terminalViewWrapper = FakeTerminalViewWrapper()
    action.terminalViewWrapper = terminalViewWrapper

    // Perform action
    myFixture.testAction(action)

    // Verify right arguments are passed into the terminal
    val expectedComand = TerminalCommand("echo Hello World", project.basePath, PROJECT_GEN_TAB_NAME)
    terminalViewWrapper.assertCommand(expectedComand)
  }

  fun testTerminalViewNotRunningWhenActionDisabled() {
    val action = ProjectGenMenuAction()
    val terminalViewWrapper = FakeTerminalViewWrapper()
    action.terminalViewWrapper = terminalViewWrapper
    project.service<SkatePluginSettings>().isProjectGenMenuActionEnabled = false

    // Perform action
    myFixture.testAction(action)

    // Verify action didn't run any terminal command
    terminalViewWrapper.assertEmptyCommand()
```

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/projectgen/ProjectGenMenuAction.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.intellij.openapi.components.service
+import com.intellij.openapi.project.Project
+import com.slack.sgp.intellij.SkatePluginSettings
+
+class ProjectGenMenuAction
+@JvmOverloads
+constructor(
+  private val terminalViewWrapper: (Project) -> TerminalViewWrapper = { project: Project ->
+    RealTerminalViewWrapper(project)
+  }
```
ZacSweers, line 29: Minor nit to make this syntax cleaner

```suggestion
  private val terminalViewWrapper: (Project) -> TerminalViewWrapper = ::RealTerminalViewWrapper
```

ZacSweers (approved):
One nit but otherwise lgtm!

PR #596, opened 2023-09-30
Make debug builds use consistent version codes/names on CI

ZacSweers:
Previously, we would use interpolations of CI build numbers and user names for version codes and version names of APKs, resulting in remote build cache misses for anything involving manifests and buildconfig. This optimizes the debug build happy path by ensuring all debug builds have the same values for both, while still using the original heuristics for any other build types.

For slack employees, you can see an example build scan comparison of where this issue arises here: https://gradle-enterprise.tinyspeck.com/c/4ywvch5rawiia/vklqqwvf3ei7s/task-inputs?cacheability=cacheable

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/gradle/ApkVersioningPlugin.kt
+++ b/slack-plugin/src/main/kotlin/slack/gradle/ApkVersioningPlugin.kt
@@ -52,32 +52,47 @@ internal class ApkVersioningPlugin : Plugin<Project> {
       val versionMinor = project.localGradleProperty("versionMinor")
       val versionPatch = project.localGradleProperty("versionPatch")
 
-      val user =
-        project.ciBuildNumber
-          .map { "" }
-          // Only provider the user if this is _not_ running on jenkins. Composition of properties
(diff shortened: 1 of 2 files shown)


slack-plugin/src/main/kotlin/slack/gradle/ApkVersioningPlugin.kt
```diff
@@ -98,14 +113,28 @@ internal class ApkVersioningPlugin : Plugin<Project> {
...
     project.configure<ApplicationAndroidComponentsExtension> {
       onVariants { variant ->
         val flavorName = variant.flavorName
 
+        val versionCodeProvider =
+          if (variant.buildType == "debug") {
+            debugVersionCodeProvider
+          } else {
+            defaultVersionCodeProvider
+          }
```
linhpha, line 131: Does this mean debug CI build will have `90009999` number from now?

ZacSweers: yep

linhpha (approved):

PR #601, opened 2023-10-04
Add regex-based linkifying support for feature flag files

arpita184:
What is this PR for?

Previously, our feature flag linkifying functionality was limited to files ending with the pattern "Feature.kt". This PR Implements support to linkify Kotlin files matching specific filename regex patterns, e.g .*Feature(s)?\\.kt$


ZacSweers (commented):
Let's make this configurable instead

arpita184:
> Let's make this configurable instead

[076bce0](https://github.com/slackhq/slack-gradle-plugin/pull/601/commits/076bce0f99f5b13f329bd861afd309f94ae99f5a)

skate-plugin/src/main/resources/messages/skateBundle.properties
```diff
@@ -8,6 +8,7 @@ skate.configuration.enableProjectGenMenuAction.title=Enable Project Gen Menu Act
 skate.configuration.enableProjectGenMenuAction.description=If enabled, shows Slack Project Generator option under Tools menu
 skate.configuration.enableFeatureFlagLinking.title=Enable Feature Flag Linking
 skate.configuration.enableFeatureFlagLinking.description=If enabled, show inspections for opening feature flag links from corresponding enum entries.
+skate.configuration.featureFlagFilePatterns.title=Provide a list of comma-separated file name patterns (e.g. Feature.kt) for feature flag checking
```
ZacSweers, line 11: Specify if it's a regex, glob, suffix, etc. judging from the code it's a comma-separated list of suffixes, which isn't obvious from this labeling. Let also doc the settings property in code. Usually we use regexes or globs for flexibility, both of which are supported by the JDK.

arpita184: [f1a16cb](https://github.com/slackhq/slack-gradle-plugin/pull/601/commits/f1a16cb08b3d32c07a33dcd3f126483c6e9b0a99)

skate-plugin/src/test/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotatorTest.kt
```diff
@@ -24,19 +24,40 @@ class FeatureFlagAnnotatorTest : BaseFeatureFlagTest() {
...
+  fun `test report symbols when comma separated list of regex file patterns`() {
+    assertThat(
+        runAnnotator(enabled = true, "TestFeatures.kt", filePattern = ".*Feature(s)?\\.kt\$")
+      )
+      .isNotEmpty()
+  }
+
+  @Test
+  fun `test return empty list when test file does not match the file pattern`() {
+    assertThat(runAnnotator(enabled = true, "TestFeatures.kt", filePattern = "Feature.kt"))
```
ZacSweers, line 47: Why does this test use a different pattern?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotator.kt
```diff
@@ -58,7 +59,10 @@ class FeatureFlagAnnotator : ExternalAnnotator<List<FeatureFlagSymbol>, List<Fea
     }
   }
 
-  private fun isKotlinFeatureFile(psiFile: PsiFile): Boolean = psiFile.name.endsWith("Feature.kt")
+  private fun isKotlinFeatureFile(psiFile: PsiFile): Boolean {
+    val filePattern = psiFile.project.featureFlagFilePattern() ?: return false
+    return psiFile.name.matches(Regex(filePattern.trim()))
```
ZacSweers, line 64: You should use the pattern exactly as saved, not trimmed.

You can also tweak it to be filePattern.toRegex().matches(psiFile.name)

ZacSweers (approved):
Couple small bits before you merge but otherwise looking good 👍

skate-plugin/src/test/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotatorTest.kt
arpita184: Because this test is to check incorrect pattern returns empty list

ZacSweers: Gotcha. The regex input should escape the period though no?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/featureflags/FeatureFlagAnnotator.kt
arpita184: Is there any harm in trim()? I want to ensure we don't break functionality for a user just because of an accidental space in the manual entry field

ZacSweers: Yes, a space is a valid regex character. You should avoid hidden modifications like this

PR #622, opened 2023-10-31
Improve Skate settings UI

linhpha:
Break setting into groups to make it easier to read.

Before:
image
After
Screenshot 2023-10-31 at 9 30 57 AM


ZacSweers (approved):
This is so much better ✨, thanks!

PR #628, opened 2023-11-02
Add support for String to Enum translators

saifc:
This adds support for generating `String` to `Enum` translators, and fixes calculating the fully qualified name of deeply nested destination classes.

### DIFF
--- a/skate-plugin/src/test/kotlin/com/slack/sgp/intellij/modeltranslator/TranslatorHelperTest.kt
+++ b/skate-plugin/src/test/kotlin/com/slack/sgp/intellij/modeltranslator/TranslatorHelperTest.kt
@@ -16,10 +16,9 @@
 package com.slack.sgp.intellij.modeltranslator
 
 import com.google.common.truth.Truth.assertThat
-import com.intellij.openapi.components.service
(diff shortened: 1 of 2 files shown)


skate-plugin/src/test/kotlin/com/slack/sgp/intellij/modeltranslator/TranslatorHelperTest.kt
```diff
@@ -130,4 +156,6 @@ class TranslatorHelperTest : LightJavaCodeInsightFixtureTestCase() {
   private fun generateBody(): KtBlockExpression? {
     return TranslatorHelper.generateBody(TranslatorHelper.extractBundle(getNamedFunction())!!)
   }
+
+  private fun skatePluginSettings() = project.service<SkatePluginSettings>()
```
ZacSweers, line 160: We have an extension for this here: https://github.com/slackhq/slack-gradle-plugin/blob/main/skate-plugin/src/main/kotlin/com/slack/sgp/intellij/util/ProjectUtils.kt#L22

ZacSweers (approved):
Thanks for the good tests on this, that made it easier to understand the goal of this. Just one nit before you merge!

PR #677, opened 2023-12-08
Integrate buildconfig plugin to project to replace old version templates work

ZacSweers:
This uses a nicer plugin to handle buildconfig constants


slack-plugin/version-templates/KotlinBuildConfig.kt
```diff
@@ -1,9 +0,0 @@
-@file:JvmName("KotlinBuildConfig")
```
linhpha, line 1: Why is this removed? Still seeing its reference here
https://github.com/slackhq/slack-gradle-plugin/pull/677/files#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06R180.

ZacSweers: That's different than this file actually, they're just named the same. That other def is here: https://github.com/slackhq/slack-gradle-plugin/pull/677/files#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06R122

build.gradle.kts
```diff
@@ -186,16 +180,18 @@ val kotlinVersion = libs.versions.kotlin.get()
...
-    .matching { it.name == "copyVersionTemplates" }
-    .configureEach {
-      val templatesMap = kotlinBuildConfig.asTemplatesMap()
-      inputs.property("buildversions", templatesMap.hashCode())
-      expand(templatesMap)
+  if (project.path == ":slack-plugin") {
+    project.pluginManager.withPlugin("com.github.gmazzo.buildconfig") {
+      configure<BuildConfigExtension> {
+        buildConfigField("String", "KOTLIN_VERSION", "\"$kotlinVersion\"")
+        // Using Any here due to https://github.com/gmazzo/gradle-buildconfig-plugin/issues/9
```
ZacSweers, line 187: PR'd a hopeful fix here https://github.com/gmazzo/gradle-buildconfig-plugin/pull/98

PR #684, opened 2023-12-13
Add code owner pane to Skate plugin

ryan-moore-slack:
This PR adds a code owner widget to the bottom right status bar of Jetbrains IDEs.

### DIFF
--- a/skate-plugin/src/main/kotlin/com/slack/sgp/intellij/ui/CodeOwnerWidget.kt
+++ b/skate-plugin/src/main/kotlin/com/slack/sgp/intellij/ui/CodeOwnerWidget.kt
@@ -1,0 +1,141 @@
+/*
+ * Copyright (C) 2023 Slack Technologies, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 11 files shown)


linhpha (approved):
🎉

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/ui/CodeOwnerWidget.kt
```diff
@@ -0,0 +1,142 @@
...
+  override fun ID() = CODE_OWNER_WIDGET_ID
+
+  override fun getPresentation(): StatusBarWidget.WidgetPresentation = this
+
+  override fun selectionChanged(event: FileEditorManagerEvent) {
+    // New selected file is taken from a few sources in the following order:
+    // 1. newFile from FileEditorManagerEvent
+    // 2. selectedFile from EditorBasedWidget superclass
+    // 3. The previously selected file cached here
+    currentSelectedFile = event.newFile
```
linhpha, line 70: Does this mean we don't cache the selected file anymore? Also it looks like the comment above should be updated?

ryan-moore-slack: yea correct, relying on the `event.newFile` seems much more reliable after some real-world use. Good catch on the comment, updated!

PR #691, opened 2023-12-18
Expose compilerOptions for compose configuration

ZacSweers:
This makes it easier to specify custom options to compose in a project.


slack-plugin/src/main/kotlin/slack/gradle/SlackExtension.kt
```diff
@@ -696,13 +700,45 @@ constructor(
   internal val enabled = objects.property<Boolean>().convention(false)
   internal val multiplatform = objects.property<Boolean>().convention(false)
 
+  /**
+   * Configures the compiler options for Compose. This is a list of strings that will be passed into
+   * the underlying kotlinc invocation. Note that you should _not_ include the plugin prefix, just
+   * the simple args directly.
```
ZacSweers, line 706:
```suggestion
   * the simple key=value options directly.
```

jstewart5000 (approved):

PR #706, opened 2024-01-04
Increase Kotlin daemon default memory percentage from `50` to `55`

jstewart5000:


ZacSweers:
@jstewart5000 since this is configurable via env flag, do we need to change the default here?

jstewart5000:
> @jstewart5000 since this is configurable via env flag, do we need to change the default here?

I can add this as env. variable in our `android_preflight.sh` if we want to leave the default to `50%` here. Let me know what you think.

ZacSweers:
yeah let's do that 👍

PR #710, opened 2024-01-09
More cleanup and fixes to artifact publishing

ZacSweers:
This does a few followup fixes to the artifact publishing we set up, focused on better supporting projects that don't publish artifacts and gracefully ignoring them in the root project when consuming them. This is accomplished via `lenient()` API in `artifactView`, which has wildly unhelpful docs but someone told me could help :). This is important for cases like `com.android.test` projects, which don't contribute unit tests or lint reports.

As a failsafe, I've also added a property to eagerly initialize all publishers in each subproject.

Also includes a couple other opportunistic fixes for the robolectric jars task path and improved common attrs for debugging.


slack-plugin/src/main/kotlin/slack/gradle/artifacts/ArtifactUtil.kt
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+package slack.gradle.artifacts
+
+import org.gradle.api.Project
+import org.gradle.api.attributes.AttributeContainer
+import org.gradle.api.attributes.Category
+import org.gradle.api.attributes.Usage
+
+internal fun AttributeContainer.addCommonAttributes(project: Project, category: String) {
```
ZacSweers, line 23: Added these to make them more obvious in debugging

jstewart5000 (approved):

PR #713, opened 2024-01-10
Fix NPE in codeowners widget

ryan-moore-slack:
Fixes a NPE when `myStatusBar` is null (assuming not yet available or disabled) in codeowners widget update code.


skate-plugin/src/main/kotlin/com/slack/sgp/intellij/ui/CodeOwnerWidget.kt
```diff
@@ -67,7 +67,9 @@ class CodeOwnerWidget(project: Project) :
     // We don't worry if it's null as that may be expected
     // and will lead to the widget displaying "none" owners.
     currentSelectedFile = event.newFile
-    myStatusBar.updateWidget(ID())
+    if (myStatusBar != null) {
+      myStatusBar.updateWidget(ID())
+    }
```
ZacSweers, line 72:
```suggestion
    myStatusBar?.updateWidget(ID())
```

ZacSweers (approved):

PR #715, opened 2024-01-11
Make PropertyResolver public + productionize API a bit

ZacSweers:
This allows use for it from other plugins and buildscripts


bryanstern (approved):
🙇

PR #744, opened 2024-02-16
Migrate Project Gen to Skate

linhpha:
This PR moves Project Gen Compose app to Skate so it can be called directly from the IDE. I made some modifications, including:

* Add a new `ProjectGenWindow` that extends `DialogWrapper` to wrap around the Compose app
* Create `projectgenlock` on the fly and delete it when dialog is dismissed
* Add an Alert dialog on file path errors
* Grab the `DarkMode` status from the IDE Theme using `JBColors.isBright`
* Make the `Quit` button to exit the dialog. It was exiting the application before
* Remove the old `TerminalWrapper` code

Remaining work I haven't been able to port over. Running into some issues that might take more time
* Updating the `Font` to `Lato`
* Adding the icon of the app


ZacSweers:
Re: the remaining work

I think we should actually try using the jetbrains/jewel compose theme instead to make this look native in the IDE rather than material. No need to add custom fonts or bring our own icon. Can be a follow up though!

Another thought: can we change the done dialog and do something like a "Close and sync" + "Close", with the former being the primary?

ZacSweers (approved):

skate-plugin/build.gradle.kts
```diff
@@ -24,6 +24,7 @@ plugins {
   alias(libs.plugins.intellij)
   alias(libs.plugins.pluginUploader)
   alias(libs.plugins.buildConfig)
+  alias(libs.plugins.compose)
```
ZacSweers, line 27: We may also need the kotlin.multiplatform plugin here in place of kotlin.jvm but not sure

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/projectgen/DesktopColors.kt
```diff
@@ -0,0 +1,153 @@
+/*
+ * Copyright (C) 2022 Slack Technologies, LLC
```
ZacSweers, line 2:
```suggestion
 * Copyright (C) 2024 Slack Technologies, LLC
```

Can you fix up the other new files too before you merge?

skate-plugin/build.gradle.kts
linhpha: I was thinking since Skate is Intellij plugin, having kotlin.jvm should be enough. Does having kotlin.multiplatform mean we'll extend to other platforms like iOS?

ZacSweers: Not necessarily, it just sets up the KMP structure. If you look at the build file in the original repo, it does the same thing.

linhpha: ah I see. I can try moving the KMP structure over

PR #748, opened 2024-02-21
Add sync action on successful project gen

linhpha:
When project got generated successfully, add an option to `Quit & Sync` that exits project gen dialog and does Gradle sync.
image

https://github.com/slackhq/slack-gradle-plugin/assets/12748234/469c74b5-baa0-4133-be51-e4820a276b06


ZacSweers (approved):
Nice! One side ask - could we update the wording to be "close" instead of "quit"? So it doesn't make it seem like it might quit android studio

PR #752, opened 2024-02-27
Fix mkdocs lists on the main page

TWiStErRob:
From main page: note the dashes, some of those are bullet points, some are descriptions of bullet points.

mkdocs's markdown doesn't like non-separate bullet lists.
Most of the project has new lines before lists, except these on the main page.

Example:
image


ZacSweers (approved):
Thanks!

PR #753, opened 2024-02-27
Add circuit to project gen option

linhpha:
Add `circuit` option to project gen. Once selected, project gen will add circuit set up to new module BUILD file:
```
  features {
    circuit()
  }
```

Screenshot:
image


ZacSweers (approved):
Nice! There's a little more available in the DSL but I think this is probably what most people use

linhpha:
> Nice! There's a little more available in the DSL but I think this is probably what most people use

Could you point me to the rest? I found there's also these 2 options `circuit(codegen = false, foundation = true)` so can add them as follow up

ZacSweers:
Yep those are the only two 👍

PR #756, opened 2024-02-28
Start tracing IDE indexing time with Skate plugin

linhpha:
Start tracing IDE indexing time with Skate plugin.

This PR adds a new listener that extends `ProjectIndexingHistoryListener` to send a trace when indexing is complete. Example data trace:

image

https://analytics.tinyspeck.com/v3/explore/1000000001254057


skate-plugin/src/main/kotlin/com/slack/sgp/intellij/tracing/SkateTraceReporter.kt
```diff
@@ -48,13 +47,12 @@ class SkateTraceReporter(private val project: Project, private val offline: Bool
...
-    } else {
-      val endPoint =
-        project.tracingEndpoint()
-          ?: throw RuntimeException("Required tracing endpoint to upload analytics")
+    val endPoint = project.tracingEndpoint()
+    if (endPoint != null) {
       SimpleTraceReporter(endPoint, loggingClient)
+    } else {
+      LOG.info("No endpoint set up for tracing")
+      NoOpTraceReporter
```
linhpha, line 55: Stop throwing exception here cause unit test code started making network request to send up the trace. The endpoint should be set up when installing the plugin but empty for unit test because we don't have skate.xml file set up in unit test environment.

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/idemetrics/IndexingListener.kt
```diff
@@ -0,0 +1,61 @@
...
+      )
+      addTag(
+        IndexingEvent.INDEXING_DURATION.name,
+        projectIndexingHistory.times.indexingDuration.toMillis(),
+      )
+      addTag(IndexingEvent.WAS_INTERRUPTED.name, projectIndexingHistory.times.wasInterrupted)
+      addTag(IndexingEvent.SCANNING_TYPE.name, projectIndexingHistory.times.scanningType.name)
+      addTag("event", IndexingEvent.INDEXING_COMPLETED.name)
+    }
+    SkateTraceReporter(project)
```
ZacSweers, line 52: Should we cache this reporter instance in a service somewhere?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/tracing/SkateSpanBuilder.kt
```diff
@@ -22,25 +22,23 @@ import com.slack.sgp.tracing.model.newTagBuilder
...
-  fun addSpanTag(key: String, value: String) {
-    keyValueList.apply { key tagTo value }
+  fun addTag(key: String, value: String) {
+    keyValueList.apply { key.lowercase() tagTo value }
   }
 
-  fun addSpanTag(key: String, value: SkateTracingEvent) {
-    keyValueList.apply { key tagTo value.type.name }
+  fun addTag(key: String, event: SkateTracingEvent) {
+    keyValueList.apply { key.lowercase() tagTo event.name }
```
ZacSweers, line 30: Let's always specify a `Locale` argument to these (i.e. `Locale.US`) so we don't have any inconsistencies across different local user locales

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/tracing/SkateTracingEvent.kt
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.slack.sgp.intellij.tracing
+
+interface SkateTracingEvent {
```
ZacSweers, line 18: For namespacing and sealedness, what if we make this `sealed interface SkateTracingEvent` and nested all the subtypes as nested classes in this without the `Event` suffix? i.e. `SkateTracingEvent.ProjectGen.PROJECT_GEN_OPEN`

ZacSweers (commented):
Sorry for some reason some comments I thought I'd included in the first pass didn't show up!

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/idemetrics/IndexingListener.kt
linhpha: That's a really good point. I haven't really thought about that. Ended up trying to initialize the instance lazily for each class here https://github.com/slackhq/slack-gradle-plugin/pull/756/commits/14c05f8d2e153ee0b8e59c5469aaffbd02d8b635. Let me know if there's a better way to write this

ZacSweers: I think that's close, but we should put it somewhere more common than the whats new window I think? Does `SkateProjectService` work?

linhpha: Tried creating `SkateTraceService` here [here](https://github.com/slackhq/slack-gradle-plugin/blob/e8ebb38959cfe626382332675c7884f4669115ae/skate-plugin/src/main/kotlin/com/slack/sgp/intellij/tracing/SkateTraceService.kt). I ended up not using lazy initialization anymore in case if someone opened another repo from the IDE, then we need to reinitialize the trace reporter

ZacSweers (approved):

PR #757, opened 2024-02-29
Migrate from csv to yaml based owner file

ryan-moore-slack:
Update code owner file parsing to new yaml format.


skate-plugin/src/main/kotlin/com/slack/sgp/intellij/SkatePluginSettings.kt
```diff
@@ -26,7 +26,7 @@ import org.jetbrains.annotations.VisibleForTesting
 internal const val DEFAULT_TRANSLATOR_SOURCE_MODELS_PACKAGE_NAME = "slack.api.schemas"
 @VisibleForTesting internal const val DEFAULT_TRANSLATOR_FILE_NAME_SUFFIX = "Translator.kt"
 @VisibleForTesting internal const val DEFAULT_TRANSLATOR_ENUM_IDENTIFIER = "name"
-internal const val DEFAULT_CODE_OWNER_FILE_PATH = "config/code-ownership/code_ownership.csv"
+internal const val DEFAULT_CODE_OWNER_FILE_PATH = "code_health_team_config.yaml"
```
ZacSweers, line 29: Let's remove this default actually (default to missing) and check in the path directly in `skate.xml` in consuming projects

ryan-moore-slack: makes sense, removed.

linhpha (approved):

PR #760, opened 2024-02-29
Set change notes when publishing Skate

linhpha:
Enable the code to set up change notes. Upon this change, planning to update the Github Action publish job to takes on a Release notes argument and write it to a `change-notes.md` file before running the publish task

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -369,7 +369,7 @@ subprojects {
           pluginId.set(pluginDetails.pluginId)
           version.set(pluginDetails.version)
           pluginDescription.set(pluginDetails.description)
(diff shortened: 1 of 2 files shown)


ZacSweers (approved):
FYI GitHub actions only ever lets you fill in one line for changes I think, so I may be hard to put meaningful notes in :/

linhpha:
> FYI GitHub actions only ever lets you fill in one line for changes I think, so I may be hard to put meaningful notes in :/

ugh yeah I tested it out and GH action is not happy with multiline strings. We can go with manual pushing the change notes before running the publishing job. I wish there is a better way.. Moving the job to Buildkite is another option. I'm pretty sure Buildkite accepts multiline strings

PR #782, opened 2024-03-16
update architecture for compose library

linhpha:
fixed https://slack-pde.slack.com/archives/C058XKY4DRV/p1710438490115959


ZacSweers:
hmm, is there a way we can make a universal binary? The focus plugin seems to accomplish this by just adding all the targets: https://github.com/DevSrSouza/dropbox-focus-intellij-plugin/blob/main/build.gradle.kts#L38-L43

ZacSweers (approved):

PR #785, opened 2024-03-18
Trace Gradle Sync with Skate

linhpha:

### DIFF
--- a/skate-plugin/build.gradle.kts
+++ b/skate-plugin/build.gradle.kts
@@ -43,6 +43,7 @@ intellij {
   plugins.add("org.intellij.plugins.markdown")
   plugins.add("org.jetbrains.plugins.terminal")
   plugins.add("org.jetbrains.kotlin")
+  plugins.add("org.jetbrains.android")
 }
 
 fun isGitHash(hash: String): Boolean {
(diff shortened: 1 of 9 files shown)


ZacSweers (approved):
Out of curiosity, why do we want to trace here when we have this in develocity too?

skate-plugin/build.gradle.kts
```diff
@@ -43,6 +43,7 @@ intellij {
   plugins.add("org.intellij.plugins.markdown")
   plugins.add("org.jetbrains.plugins.terminal")
   plugins.add("org.jetbrains.kotlin")
+  plugins.add("org.jetbrains.android")
```
ZacSweers, line 46: I'm surprised we don't need the google-made one. Are these the same?

linhpha:
> Out of curiosity, why do we want to trace here when we have this in develocity too?

I think it'd be nice to compare the data here vs Develocity to see if there's any difference. Another nice extra just brought up by Jason is we can get from this data is when someone skipped sync. I think it'd be a deal breaker if adding tracing makes the IDE performance worse. So far, I think it's been quite fast so I want to go with it and see how the data looks like. Lmk if you have other thoughts

skate-plugin/build.gradle.kts
linhpha: Which google-made plugin you're referring to? I'm adding this one because the SyncListener is part of if
https://github.com/JetBrains/android/tree/master

PR #799, opened 2024-04-02
Create Circuit feature file template from Skate

linhpha:
* This PR adds `Circuit Feature` generation to Skate. Intead of using YAML File template within the IDE, I'm generating the feature template with Kotlin Poet. The reason for this move is we've recently migrated `Project Generator` to be a separate module in `sgp`. This helps remove the need for `circuit gen` to depend on Intellij plugin API so we can integrate it into `project gen` easier. 

* Some other changes:

    - Simplify the generated code to create more stubs
    - Add ability to add missing Gradle dependencies
    - Make `assisted injection` optional
    - Make test generation optional
   
Here's what the screen looks like:
image


ZacSweers (commented):
Few high level notes before we get into the nitty gritty reviews

- Can we use kotlinpoet to generate these kotlin files instead?
- We should just generate stubs. Generating these full counter implementations or sample tests is going to result in people spending a chunk of time deleting them (or worse, checking them in)
- Can we make the assisted injection bits optional? Many screens don't need these both injected
- Can we make as much of the hardcoded custom slack test infra bits configurable?
- Can we try to stick to using `java.nio.path.*` Path APIs instead of `File`? The former is a more modern replacement.
- Can we add necessary dependencies to the build file if they're missing?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+    val circuitUiState = ClassName("com.slack.circuit.runtime", "CircuitUiState")
+    val screenInterface = ClassName("com.slack.circuit.runtime.screen", "Screen")
+    val eventInterface =
+      TypeSpec.interfaceBuilder("Event")
+        .addAnnotation(Immutable::class)
+        .addModifiers(KModifier.SEALED)
+        .addSuperinterface(circuitUiEvent)
+        .build()
+
+    val eventInterfaceName = eventInterface.name?.let { ClassName("", it) }
```
ZacSweers, line 74: Shouldn't the package name be the package name of the outer screen? You can do something like this

```kotlin
val screen = ClassName(<package>, <screen name>)
val eventInterfaceCN = screen.nested("Event")
val eventInterface = TypeSpec.interfaceBuilder(eventInterfaceCN.simpleName)
    // ...
```

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+                    receiver = eventInterfaceName,
+                    returnType = Unit::class.asTypeName(),
+                  ),
+                )
+                .defaultValue("{}")
+                .build()
+            )
+            .build()
+        )
+        .addProperty(PropertySpec.builder("message", String::class).initializer("message").build())
```
ZacSweers, line 96: Let's not include this property, for the same reason as other dummy data - people will check it in :)

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+    return FileSpec.builder(packageName, circuitScreenClass)
+      .addType(
+        TypeSpec.classBuilder(circuitScreenClass)
+          .addSuperinterface(screenInterface)
+          .addAnnotation(parcelize)
+          .addType(stateClass)
+          .addType(eventInterface)
+          .build()
+      )
+      .build()
```
ZacSweers, line 120: There's a shortcut where you can pass the typespec directly as a param to `FileSpec.create()` instead

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+     }
+    */
+    val screenClassName = "${className}Screen"
+    val presenterClassName = "${className}Presenter"
+    val assistedInject = ClassName("dagger.assisted", "AssistedInject")
+    val assisted = ClassName("dagger.assisted", "Assisted")
+    val assistedFactory = ClassName("dagger.assisted", "AssistedFactory")
+
+    val factoryBuilder =
+      TypeSpec.interfaceBuilder("Factory").apply {
```
ZacSweers, line 165: Let's make these factory interfaces `fun interface`

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+            .apply {
+              if (useAssistedInjection) {
+                addAnnotation(assisted)
+              }
+            }
+            .build()
+        )
+        addParameter(
+          ParameterSpec.builder("navigator", ClassName("com.slack.circuit.runtime", "Navigator"))
+            .apply {
```
ZacSweers, line 187: Can we let the user choose which params to assist? Then we can pass in a list of assisted parameters (and an empty list would imply that `useAssistedInjection = false`)

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+        }
+      }
+
+    val constructorBuilder =
+      FunSpec.constructorBuilder().apply {
+        if (useAssistedInjection) {
+          addAnnotation(assistedInject)
+        }
+        addParameter(
+          ParameterSpec.builder("screen", ClassName("", screenClassName))
```
ZacSweers, line 177: This will end up in a ClassName with no package. Can we pass in the original `Screen` classname directly?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+            .apply {
+              if (useAssistedInjection) {
+                addAnnotation(assisted)
+              }
+            }
+            .build()
+        )
+        addParameter(
+          ParameterSpec.builder("navigator", ClassName("com.slack.circuit.runtime", "Navigator"))
```
ZacSweers, line 186: Let's extract `Navigator` to a top-level constant

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+            """
+         TODO (remove): This Circuit [Presenter] was generated without UI or [CircuitInject], and can be
+         used with legacy views via `by circuitState()` in your Fragment or Activity.
+          """
+              .trimIndent()
+          )
+        }
+        primaryConstructor(constructorBuilder.build())
+          .addSuperinterface(
+            ClassName("com.slack.circuit.runtime.presenter", "Presenter")
```
ZacSweers, line 209: Another one to extract top-level

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+              .parameterizedBy(ClassName(packageName, "${className}Screen.State"))
+          )
+          .addProperty(
+            PropertySpec.builder("screen", ClassName("", "${className}Screen"))
```
ZacSweers, line 213: Another empty package

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+              .addModifiers(KModifier.PRIVATE)
```
ZacSweers, line 214: These modifiers will get stripped by your post-processing to remove `public `. Can we refine it to be more specific in which declarations it strips the prefix from?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+         TODO (remove): This Circuit [Presenter] was generated without UI or [CircuitInject], and can be
+         used with legacy views via `by circuitState()` in your Fragment or Activity.
+          """
+              .trimIndent()
+          )
+        }
+        primaryConstructor(constructorBuilder.build())
+          .addSuperinterface(
+            ClassName("com.slack.circuit.runtime.presenter", "Presenter")
+              .parameterizedBy(ClassName(packageName, "${className}Screen.State"))
```
ZacSweers, line 210:
```suggestion
              .parameterizedBy(ClassName(packageName, "${className}Screen").nested("State"))
```

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+            PropertySpec.builder("navigator", ClassName("com.slack.circuit.runtime", "Navigator"))
+              .addModifiers(KModifier.PRIVATE)
+              .initializer("navigator")
+              .build()
+          )
+          .addFunction(
+            FunSpec.builder("present")
+              .addModifiers(KModifier.OVERRIDE)
+              .addAnnotation(ClassName("androidx.compose.runtime", "Composable"))
+              .returns(ClassName("", "${className}Screen.State"))
```
ZacSweers, line 228: Another empty package

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+              .addCode(
+                """
+              val scope = rememberStableCoroutineScope()
+              return ${className}Screen.State() { event ->
+              }
+            """
+                  .trimIndent()
+              )
```
ZacSweers, line 236: Any reason we're generating this scope var? Instead I'm thinking we really just want to do this

```suggestion
              .addStatement("TODO(%S)", "Implement me!")
```

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+              .build()
+          )
+          .addType(
+            factoryBuilder
+              .addAnnotation(
+                AnnotationSpec.builder(
+                    ClassName("com.slack.circuit.codegen.annotations", "CircuitInject")
```
ZacSweers, line 243: Extract a constant

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+                  )
+                  .addMember("%T::class", ClassName(packageName, "${className}Screen"))
+                  .addMember("%T::class", ClassName("slack.di", "UserScope"))
```
ZacSweers, line 246: Can we make this configurable?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+              )
+              .build()
+          )
+          .addType(
+            factoryBuilder
+              .addAnnotation(
+                AnnotationSpec.builder(
+                    ClassName("com.slack.circuit.codegen.annotations", "CircuitInject")
+                  )
+                  .addMember("%T::class", ClassName(packageName, "${className}Screen"))
```
ZacSweers, line 245: Definitely think you want to extract a `ClassName(packageName, "${className}Screen")` variable that you can reuse

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+              val scope = rememberStableCoroutineScope()
+              return ${className}Screen.State() { event ->
+              }
+            """
+                  .trimIndent()
+              )
+              .build()
+          )
+          .addType(
+            factoryBuilder
```
ZacSweers, line 240: This should be gated on assisted injection right?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+    )
+    @Composable
+    fun Feature(state: ggScreen.State, modifier: Modifier = Modifier) {
+    }
+     */
+    val uiFunction =
+      FunSpec.builder(className)
+        .addAnnotation(
+          AnnotationSpec.builder(
+              ClassName("com.slack.circuit.codegen.annotations", "CircuitInject")
```
ZacSweers, line 281: Extract a constant

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+  override val fileSuffix = ""
+
+  override fun generate(packageName: String, className: String): FileSpec {
+    /*
+    @CircuitInject(
+      FeatureScreen::class,
+      UserScope::class,
+    )
+    @Composable
+    fun Feature(state: ggScreen.State, modifier: Modifier = Modifier) {
```
ZacSweers, line 274: Is this right?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+            .addMember("%T::class", ClassName(packageName, "${className}Screen"))
+            .addMember("%T::class", ClassName("slack.di", "UserScope"))
+            .build()
+        )
+        .addAnnotation(Composable::class)
+        .addParameter(
+          ParameterSpec.builder("state", ClassName(packageName, "${className}Screen.State")).build()
+        )
+        .addParameter(
+          ParameterSpec.builder("modifier", ClassName("androidx.compose.ui", "Modifier"))
```
ZacSweers, line 292: Extract a constant for Modifier

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+    val testClassSpec =
+      TypeSpec.classBuilder("${className}${fileSuffix}")
+        .apply {
+          if (baseClass != null) {
+            superclass(ClassName.bestGuess(baseClass))
+          }
+        }
+        .build()
+
+    return FileSpec.builder(packageName, "${className}${fileSuffix}").addType(testClassSpec).build()
```
ZacSweers, line 320: Same here re: `FileSpec.get(...)` use to help

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/circuitgenerator/CircuitFeatureGenerator.kt
```diff
@@ -0,0 +1,110 @@
...
+  override fun actionPerformed(event: AnActionEvent) {
+    val currentProject = event.project ?: return
+    if (!currentProject.isCircuitGeneratorEnabled()) return
+    val selectedFile = event.getData(CommonDataKeys.VIRTUAL_FILE) ?: return
+    val selectedDirectory =
+      if (selectedFile.isDirectory) selectedFile.path else selectedFile.parent.path
+    showFeatureDialog(currentProject, selectedDirectory)
+  }
+
+  private fun showFeatureDialog(project: Project, directory: String) {
```
ZacSweers, line 51: Can we write this UI with compose + jewel?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/circuitgenerator/UdfViewModelGenerator.kt
```diff
@@ -0,0 +1,87 @@
...
+import com.intellij.ui.dsl.builder.bindSelected
+import com.intellij.ui.dsl.builder.bindText
+import com.intellij.ui.dsl.builder.panel
+import com.slack.sgp.intellij.util.getJavaPackageName
+import com.slack.sgp.intellij.util.isCircuitGeneratorEnabled
+import java.io.File
+import slack.tooling.projectgen.circuitgen.CircuitPresenter
+import slack.tooling.projectgen.circuitgen.CircuitScreen
+
+class UdfViewModelGenerator : AnAction(), DumbAware {
```
ZacSweers, line 36: Are we adding component generation for UdfViewModel in this PR too?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/util/StringExtensions.kt
```diff
@@ -20,3 +20,11 @@ import com.google.common.base.CaseFormat
 fun String.snakeToCamelCase(): String {
   return CaseFormat.LOWER_UNDERSCORE.to(CaseFormat.LOWER_CAMEL, this)
 }
+
+fun String.getJavaPackageName(): String? {
```
ZacSweers, line 24: This feels tricky to be putting on string. Presumably we usually have a File or Path, could we use that and then get a concrete list of parent dirs and relativize them to the original `src` source set?

ZacSweers (commented):
I think this is missing logic & tests for non-assisted-inject presenters

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/util/StringExtensions.kt
linhpha: Moved this to `PathsExtension` here https://github.com/slackhq/slack-gradle-plugin/pull/799/commits/6986e693d5d0bf42d69c263050612fb0634dbdbc#diff-2ce36a3822d274d7a4642aba1324108f9e61db2fe992d7f11aabf443147cce97R20

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/circuitgenerator/UdfViewModelGenerator.kt
linhpha: Oops, yup i'm adding UdfViewModel as well. Updated in https://github.com/slackhq/slack-gradle-plugin/pull/799/commits/6986e693d5d0bf42d69c263050612fb0634dbdbc

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
linhpha: Added a `CircuitGenClassNames` class

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
linhpha: Updated the UI to have options for user to choose list of assisted param
image

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
linhpha: Just want to clarify, I'm having 2 options for adding assisted injection and circuit injection. Should the circuit injection dependent on assisted injection?
image

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
linhpha: Fixed :)

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,322 @@
...
+
+  fun isTestComponent(): Boolean = false
+
+  fun writeToFile(directory: String, packageName: String?, className: String) {
+    val baseDirectory =
+      if (isTestComponent()) directory.replace("src/main", "src/test") else directory
+    val fileSpec = generate(packageName.orEmpty(), className)
+    File(baseDirectory).apply { if (!exists()) mkdirs() }
+    val stringWriter = StringWriter()
+    fileSpec.writeTo(stringWriter)
```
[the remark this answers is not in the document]
linhpha: I found it a bit more complicated cause when user right click on the folder to add new feature, the dir can contain the package name already. So `fileSpec.writeTo(File)` can create more nested package names than needed. Here's what I tried with your suggestion but ended up reverting back cause it felt more tricky. Lmk what you think

```
val baseDirectory =
  if (isTestComponent()) Path.of(directory.pathString.replace("src/main", "src/test")) else directory
val fileSpec = generate(packageName.orEmpty(), "$className$fileSuffix")
Files.createDirectories(baseDirectory)

val trimmedPath = packageName?.let {
  val packagePath = it.replace('.', '/')
  Path.of(baseDirectory.toString().replace(packagePath, ""))
} ?: baseDirectory
fileSpec.writeTo(trimmedPath)

val filePath = baseDirectory.resolve("$className$fileSuffix.kt")
val generatedCode = Files.readString(filePath)
val updatedCode = generatedCode.replace("public ", "")
Files.writeString(filePath, updatedCode)
```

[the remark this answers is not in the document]
ZacSweers: In my head, I expect all you'd need are

```kotlin
fun generateFile(projectDir: Path, selectedDir: Path) {
  // Get the path of the src/.../kotlin dir
  val srcDir = generateSequence(selectedDir) { it.parent }.find { it.name == "kotlin" || it.name == "java")
  val packageName = selectedDir.relativeTo(srcDir).replace(File.separatorChar, ".")
  val spec = createFileSpec(packageName)
  spec.writeTo(srcDir)
}
```

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/UiElement.kt
```diff
@@ -35,15 +35,19 @@ internal object DividerElement : UiElement {
...
+) : UiElement {
   override var isVisible: Boolean = true
 }
 
 internal class CheckboxElement(
   private val initialValue: Boolean,
   val name: String,
   val hint: String,
-  val indentLevel: Int = 0,
+  var indentLevel: Int = 0,
```
ZacSweers, line 50: If you want this to be mutable then we should make it a `State` variable like `isChecked` is below. Ideally we shouldn't need this though? Do we have checkboxes whose' indention is changing conditionally after creation?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,408 @@
...
+class CircuitScreen : CircuitComponent {
+  override val fileSuffix = "Screen"
+
+  override fun generate(packageName: String, className: String): FileSpec {
+    /**
+     * Generate Circuit Screen class, eg:
+     *
+     * @Parcelize class ${NAME}Screen : Screen { data class State( val eventSink: Event.() -> Unit =
+     *   {}, ) : CircuitUiState
+     * @Immutable sealed interface Event : CircuitUiEvent }
```
ZacSweers, line 85: Let's put this in code blocks so they're not reformatted. Same for other cases

```suggestion
     * ```kotlin
     * @Parcelize class ${NAME}Screen : Screen { data class State( val eventSink: Event.() -> Unit =
     *   {}, ) : CircuitUiState
     * @Immutable sealed interface Event : CircuitUiEvent }
     * ```
```

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,408 @@
...
+  private val additionalCircuitInject: Set<ClassName> = setOf(),
+  private val noUi: Boolean = false,
+) : CircuitComponent {
+  override val fileSuffix = "Presenter"
+
+  override fun generate(packageName: String, className: String): FileSpec {
+    /**
+     * Generate Circuit Presenter class, eg: class ${NAME}Presenter @AssistedInject constructor(
+     *
+     * @Assisted private val screen: ${NAME}Screen,
```
ZacSweers, line 146: Same here with formatting

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,408 @@
...
+        .addType(stateClass)
+        .addType(eventInterface)
+        .build()
+    return FileSpec.get(packageName, typeSpec)
+  }
+}
+
+class CircuitPresenter(
+  private val assistedInjection: AssistedInjectionConfig,
+  private val additionalCircuitInject: Set<ClassName> = setOf(),
```
ZacSweers, line 137: Since the order matters, let's use a List

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,408 @@
...
+  private val noUi: Boolean = false,
```
ZacSweers, line 138: Instead of a double-negative, let's just have `ui: Boolean = true`

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,408 @@
...
+    val factoryBuilder =
+      TypeSpec.funInterfaceBuilder("Factory").apply {
+        if (addAssistedInjection) {
+          addAnnotation(assistedFactory)
+        }
+        addAnnotation(
+          AnnotationSpec.builder(CIRCUIT_INJECT)
+            .apply {
+              addMember("%T::class", screenClass)
+              additionalCircuitInject.forEach { inject -> addMember("%T::class", inject) }
```
ZacSweers, line 168: CircuitInject accepts exactly two args, I think you got something mixed up here?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
```diff
@@ -0,0 +1,408 @@
...
+        .addFunction(stateFun)
+        .build()
+    return FileSpec.builder(packageName, className)
+      .addImport(closeableCoroutineScope.packageName, closeableCoroutineScope.simpleName)
+      .addType(viewModelClass)
+      .build()
+  }
+}
+
+data class AssistedInjectionConfig(val screen: Boolean = false, val navigator: Boolean = false)
```
ZacSweers, line 408: nit: in circuit's code gen we just model this with a list of `ParameterSpec`s, so it's extensible for future params we add too.

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitGenClassName.kt
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package slack.tooling.projectgen.circuitgen
+
+import com.squareup.kotlinpoet.ClassName
+
+class CircuitGenClassNames {
+  companion object {
```
ZacSweers, line 21: You can just make `CircuitGenClassNames` an object

```suggestion
object CircuitGenClassNames {
```

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitGenPresenter.kt
```diff
@@ -0,0 +1,236 @@
...
+      circuitUi,
+      circuitPresenter,
+      assistedInjection,
+      assistedScreen,
+      assistedNavigator,
+      circuitInject,
+      userScopeCircuitInject,
+      appScopeCircuitInject,
+      test,
+    )
```
ZacSweers, line 92: I wonder if we should make a little DSL for this kind of design as a followup

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitGenPresenter.kt
```diff
@@ -0,0 +1,236 @@
...
+  private val viewModelElements =
```
ZacSweers, line 93: I don't think we should support viewmodel generation here. It's not useful outside of slack's codebase and also not something we're recommending people write new ones of anyway

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitGenUi.kt
```diff
@@ -0,0 +1,181 @@
...
+        }
+        .build()
+    }
+    render { CircuitContent(CircuitGenScreen, circuit = circuit) }
+  }
+}
+
+@Composable
+internal fun CircuitGen(state: CircuitGenScreen.State, modifier: Modifier = Modifier) {
+
```
ZacSweers, line 121: Any reason for this space?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitGenUi.kt
```diff
@@ -0,0 +1,181 @@
...
+  Box(modifier.fillMaxSize().background(JewelTheme.globalColors.paneBackground)) {
```
ZacSweers, line 122: Is the manual background() necessary here? I assume intellij provides a default one in the panel window

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitGenUi.kt
```diff
@@ -0,0 +1,181 @@
...
+    val listState = rememberLazyListState()
+    Column(Modifier.padding(16.dp)) {
+      LazyColumn(
+        modifier = Modifier.weight(1f),
+        state = listState,
+        verticalArrangement = Arrangement.spacedBy(16.dp),
```
ZacSweers, line 128: nit: static import `spacedBy` for readability

skate-plugin/project-gen/src/jvmTest/kotlin/slack/tooling/projectgen/CircuitComponentTest.kt
```diff
@@ -0,0 +1,207 @@
...
+          TODO("Implement me!")
+        }
+
+        @CircuitInject(
+          FooScreen::class,
+          UserScope::class,
+        )
+        public fun interface Factory {
+          public fun create(screen: FooScreen, navigator: Navigator): FooPresenter
+        }
```
ZacSweers, line 93: If there's no assisted parameters, we should not generate a factory nor add the screen/navigator params. We should instead
- Put the CircuitInject annotation on the presenter class itself
- Create a primary constructor annotated with `@Inject`, i.e. `@Inject constructor()`.

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/circuitgenerator/CircuitFeatureGenerator.kt
```diff
@@ -0,0 +1,62 @@
...
+    val listener =
+      object : FileGenerationListener {
+        override fun onFilesGenerated(fileNames: String) {
+          LocalFileSystem.getInstance().refreshAndFindFileByIoFile(File(fileNames))?.let {
+            FileEditorManager.getInstance(project).openFile(it)
+          }
+          LocalFileSystem.getInstance().refresh(true)
```
ZacSweers, line 50: Do we need to refresh twice?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/circuitgenerator/UdfViewModelGenerator.kt
ZacSweers: Sorry I meant this as sort of a rhetorical question, I don't think we should be 😅

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/UiElement.kt
linhpha: Updated back to `val` https://github.com/slackhq/slack-gradle-plugin/pull/799/commits/742142fe1bbe3cc3e2d96e29383f4c4b2b6ddab9

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
linhpha: updated in https://github.com/slackhq/slack-gradle-plugin/pull/799/commits/790aff0b03a4d5bc0e12a2ed4ee3f6cf5dad71a3

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
linhpha: updated https://github.com/slackhq/slack-gradle-plugin/pull/799/commits/9dea9a05fa600d34b4b1a8c8fe12f426493136ce

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
linhpha: updated https://github.com/slackhq/slack-gradle-plugin/pull/799/commits/b695d14d10124b133737548dde5d8e1449199c74

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
linhpha: Thanks for catching that. Updated https://github.com/slackhq/slack-gradle-plugin/pull/799/commits/41e015de2bb22f65c68de702f19c193679d02324

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitGenClassName.kt
linhpha: https://github.com/slackhq/slack-gradle-plugin/pull/799/commits/70b29708ba841e1850dd82a7995675fa598ae83b

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitGenPresenter.kt
linhpha: removed ViewModel generation from this PR https://github.com/slackhq/slack-gradle-plugin/pull/799/commits/25eaeafd2f23ad70c112008154821f4ab7c6420b

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitGenUi.kt
linhpha: removed!

skate-plugin/project-gen/src/jvmTest/kotlin/slack/tooling/projectgen/CircuitComponentTest.kt
linhpha: added the case when there's no assisted params
https://github.com/slackhq/slack-gradle-plugin/pull/799/commits/0e7541ab15a1b946d3175ad56082a28ccbbf8eda

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/circuitgenerator/CircuitFeatureGenerator.kt
linhpha: I think we do. I tried to remove the second refresh but the IDE doesn't load up new files created. Line 47-49 just to get the new screen file and open it up. ` LocalFileSystem.getInstance().refresh(true)` should load the rest.

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/circuitgen/CircuitComponent.kt
linhpha: tried to make this but might not make it in time before you left. mind if i do it as a follow up?

PR #816, opened 2024-04-12
Fix compatibility issue with newer IDEs and some UI improvement to Project Gen

linhpha:
1. Ran the `runPluginVerifier` [task](https://plugins.jetbrains.com/docs/intellij/verifying-plugin-compatibility.html) locally and found one more compatibility issue for Skate with newer IDEs. This fixes the issue by adding the java module dependencies to Skate

image

As a follow up, we can add this task to CI so it can detect these type of issues earlier. 

2. Fix the alerts dialog for Project Generator. 
before
image

after

image

### DIFF
--- a/skate-plugin/change-notes.html
+++ b/skate-plugin/change-notes.html
@@ -1,4 +1,9 @@
-<![CDATA[
+<h2>0.7.5</h2>
+<ul>
+    <li>Adopt Jewel for Project Generator</li>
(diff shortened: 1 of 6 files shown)


skate-plugin/change-notes.html
ZacSweers, line 1: 🚀

ZacSweers (approved):

PR #825, opened 2024-04-25
Switch to Compose-based markdown rendering solution

ZacSweers:
This uses a multiplatform compose markdown renderer for markdown rendering, offering a nicer UI for the whatsnew panel since Studio doesn't seem likely to ever support JCEF.

In the process of this, I also made a playground project where we can test compose UIs like this and project gen without needing to run a whole IDE flow to test, to make it easier to iterate. Now it's as easy as just running the `main` functions in those files.

There are still some features missing from the markdown library like strikethrough, dividers, tables, etc, but the maintainer is responsive and I've been filing issues. This builds a best-effort bridge theme based off of Jewel. In the future we may also consider their in-development impl.

| Light | Dark |
| ----- | ---- |
| Screenshot 2024-04-25 at 2 23 59 PM | Screenshot 2024-04-25 at 2 24 01 PM |

### DIFF
--- a/skate-plugin/compose-playground/src/jvmMain/kotlin/slack/tooling/compose/playground/ProjectGenPlayground.kt
+++ b/skate-plugin/compose-playground/src/jvmMain/kotlin/slack/tooling/compose/playground/ProjectGenPlayground.kt
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C) 2024 Slack Technologies, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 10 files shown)


skate-plugin/compose-playground/src/jvmMain/kotlin/slack/tooling/compose/playground/ProjectGenPlayground.kt
```diff
@@ -0,0 +1,57 @@
...
+            }
+
+            override fun dismissDialogAndSync() {
+              println("dismissDialogAndSync")
+            }
+          },
+      )
+    }
+  }
+}
```
linhpha, line 57: this is awesome 🎉

linhpha (approved):
:shipit:

PR #827, opened 2024-04-25
Add new flag + logic to relocate project build directories

ZacSweers:
We were tipped off by our friends at another company that this helps with IDE performance immensely by completely preventing the IDE from indexing build dirs (which tend to be littered with tons of intermediate files, even if untracked). We want to experiment with this flag and see if it improves things for us too.

We might have some plugins that still hardcode lookups to `build` dirs, but we can iterate from here.

### DIFF
--- a/slack-plugin/src/main/kotlin/slack/gradle/SlackBasePlugin.kt
+++ b/slack-plugin/src/main/kotlin/slack/gradle/SlackBasePlugin.kt
@@ -42,6 +42,15 @@ internal class SlackBasePlugin @Inject constructor(private val buildFeatures: Bu
   override fun apply(target: Project) {
     val slackProperties = SlackProperties(target)
 
+    if (slackProperties.relocateBuildDir && !target.isSyncing) {
+      // <root-dir>/../<root-dir-name>-out/<project's relative path>
+      val rootDir = target.rootDir
+      val rootDirName = rootDir.name
+      target.layout.buildDirectory.set(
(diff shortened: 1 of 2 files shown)


jhperezd (approved):
👍

linhpha (approved):

SimonMarquis:
👋 Hi @ZacSweers, were you able to make it work properly on Android Studio?
I've tested it, and it seems like Android Studio does not like it, at all.
In the project panel, it shows 2 entries for the root project.
The first entry contains all modules in a flat hierarchy (which makes it hard to distinguish public/impl splitted modules), as well as all sourceSets that contains generated code (e.g. ksp, hilt, etc.)
And unfortunately, auto-navigation will always show the file in the first group.

Screenshot 2024-07-01 at 11 17 40

ZacSweers:
It's been working for us. You might need to do a fresh project import

SimonMarquis:
I retried with nowinandroid project on macOS and Windows, even after deleting `.idea` directory and importing a clean project 🤔
After opening the project the first time (so, no compilation task has been run), both Android view and Project view seems good:
Screenshot 2024-07-01 203544
Screenshot 2024-07-01 203610

But after running any compilation, the Project view shows the external directory:
Screenshot 2024-07-01 204038

Are you also using the "project" view?

ZacSweers:
Yes we use that too. I don't remember if this was fully rolled out at slack before I started my sabbatical, but in my testing at least it worked fine. This issue tracker is probably not the best place to pursue this though

PR #849, opened 2024-05-16
Fix compose compiler update warning

linhpha:
When update compose compiler plugin to 1.5.14, we're running into this error:
```
w:  `suppressKotlinVersionCompatibilityCheck` is set to the same version of Kotlin that the Compose Compiler was already expecting (Kotlin 1.9.24), and thus has no effect and should be removed.
```
PR removes the `suppressKotlinVersionCompatibilityCheck` to fix the warning. I tested by building the app locally


ZacSweers (changes requested):
@linhpha below the compose compiler version def in libs.versions.toml there's a key to specify which kotlin version it targets, we should update that instead (in this case to 1.9.24). This'll all go away in K2 anyway though fortunately

PR #850, opened 2024-05-17
Update codeowners to support new config format

jhperezd:
The `path` objects for the config are changing, this PR introduces support for the new format.

Currently `paths` looks as follows:
```
    paths:
      - app/folder1/.*
      - app/folder2/.*
      - app/folder3/subfolder/.*
```

We are adding a new `notify` flag which means that a `path` can be either a string (i.e. previous format) or an object (new format:
```
    paths:
      - path: app/folder1/.*
        notify: false
      - app/folder2/.*
      - app/folder3/subfolder/.*
```

This PR introduces a custom deserializer and a Path object to be able to handle the above. The changes are backwards compatible so old configs will still work as usual.

Testing:
- The relevant tests were updated to cover the new config
- I was able to confirm with a local build of SKATE that the new format (and the old format) are working okay


linhpha (approved):

ryan-moore-slack (approved):
Looks good! Thanks for taking care of this with the config file updates!

PR #927, opened 2024-08-21
Initial UI Window for DevXP AI with Compose 

kateliu20:
This is a very basic start to the DevXP AI integration with Skate, using Jetpack Compose for the UI. I added the Compose dependencies to Skate as they weren't there before and was able to get a basic project structure going. So far, I added just a simple UI component with the user input text field, where the "Send" and "Clear Chat" buttons just clear the text conversation. Here is a video demo:

https://github.com/user-attachments/assets/5222c43c-fab2-4066-8000-0ec476b76414

here are the updated changes based on the discussion in this PR:

https://github.com/user-attachments/assets/136b0fe6-598c-4727-ad9d-9412db7fd183

I also took changes from [Linh's changes](https://github.com/slackhq/slack-gradle-plugin/compare/lp_slack_day_devxp_api?expand=1) for the side tool window implementation.


skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
```diff
@@ -0,0 +1,69 @@
...
+  Row(modifier) {
+    TextField(
+      value = textValue,
+      onValueChange = { newText -> textValue = newText },
+      label = { Text("Start your conversation") },
+      modifier = Modifier.weight(1f),
+    )
+    Column() {
+      Box(modifier = Modifier.clickable {}.defaultMinSize(minWidth = 56.dp).padding(8.dp)) {
+        Text("Send", color = MaterialTheme.colorScheme.surface)
```
linhpha, line 57: Can we use `SwingBridgeTheme` instead so it looks more native to Intellij plugin? You can find the example of it using here https://github.com/slackhq/slack-gradle-plugin/blob/main/skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/projectgen/SlackDesktopTheme.kt#L24

gradle/libs.versions.toml
```diff
@@ -38,6 +38,7 @@ wire = "5.0.0"
 bestPracticesPlugin = { id = "com.autonomousapps.plugin-best-practices-plugin", version = "0.10" }
 buildConfig = { id = "com.github.gmazzo.buildconfig", version = "5.3.5" }
 compose = { id = "org.jetbrains.compose", version.ref = "compose-jb" }
+compose-compiler = { id = "org.jetbrains.kotlin.plugin.compose", version.ref = "kotlin" }
```
linhpha, line 41: Looks like this doesn't get used. Can we remove it?

kateliu20: Yes, I removed it.

skate-plugin/build.gradle.kts
```diff
@@ -26,6 +26,8 @@ plugins {
   alias(libs.plugins.pluginUploader)
   alias(libs.plugins.buildConfig)
   alias(libs.plugins.lint)
+  alias(libs.plugins.compose)
+  alias(libs.plugins.kotlin.plugin.compose)
```
ZacSweers, line 30: let's move this to a separate subproject like we did with the project generator, we had issues in the past getting this to stay working nice. You can add it to the project generator too and just give the project a more generic name, like skate-ui or something?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
```diff
@@ -0,0 +1,73 @@
...
+fun ConversationField(modifier: Modifier = Modifier) {
+  var textValue by remember { mutableStateOf(TextFieldValue("")) }
+  Row(modifier) {
+    TextField(
+      value = textValue,
+      onValueChange = { newText -> textValue = newText },
+      modifier = Modifier.weight(1f).padding(4.dp).height(56.dp),
+      placeholder = { Text("Start your conversation") },
+    )
+    Column() {
```
ZacSweers, line 61: Any reason for keeping these parens?

```suggestion
    Column {
```

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.jetbrains.jewel.ui.component.DefaultButton
+import org.jetbrains.jewel.ui.component.Text
+import org.jetbrains.jewel.ui.component.TextField
+
+@Composable
+fun ChatWindowCompose(modifier: Modifier = Modifier) {
+  Column(
+    modifier = modifier.fillMaxSize().background(JewelTheme.globalColors.paneBackground),
+    verticalArrangement = Arrangement.Bottom,
+  ) {
```
ZacSweers, line 46: Is the column here to save slots for adding more content later?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
```diff
@@ -0,0 +1,73 @@
...
+}
+
+@Composable
+fun ConversationField(modifier: Modifier = Modifier) {
+  var textValue by remember { mutableStateOf(TextFieldValue("")) }
+  Row(modifier) {
+    TextField(
+      value = textValue,
+      onValueChange = { newText -> textValue = newText },
+      modifier = Modifier.weight(1f).padding(4.dp).height(56.dp),
```
ZacSweers, line 58: It's better to add padding and then let your text style's font determine the height, this way people with larger fonts (for example - accessibility reasons) don't see their text cut off.

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
```diff
@@ -0,0 +1,73 @@
...
+    Column() {
+      DefaultButton(
+        modifier = Modifier.defaultMinSize(minWidth = 56.dp).padding(4.dp),
+        onClick = { textValue = TextFieldValue("") },
+      ) {
+        Text("Send")
+      }
+      Box(modifier = Modifier.clickable { textValue = TextFieldValue("") }.padding(4.dp)) {
+        Text("Clear Chat")
+      }
```
ZacSweers, line 70: Let's use an `IconButton` instead? Maybe with a trash icon?

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
```diff
@@ -0,0 +1,73 @@
...
+    TextField(
+      value = textValue,
+      onValueChange = { newText -> textValue = newText },
+      modifier = Modifier.weight(1f).padding(4.dp).height(56.dp),
+      placeholder = { Text("Start your conversation") },
+    )
+    Column() {
+      DefaultButton(
+        modifier = Modifier.defaultMinSize(minWidth = 56.dp).padding(4.dp),
+        onClick = { textValue = TextFieldValue("") },
```
ZacSweers, line 64: Is there a way to clear the existing value instead?

ZacSweers (commented):
FYI JB open sourced a bunch of IJ icons we can reuse https://intellij-icons.jetbrains.design/

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
kateliu20: Yeah, I also wanted it so that the arrangement of the text field to be at the bottom, and then I plan to add more content after into the column.

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
kateliu20: Thanks for bringing that up, I didn't consider that.

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
kateliu20: I thought we could just set it up for if I wanted to add parameters in the future, but I will delete them for now.

ZacSweers: yep just add them when we need 'em 👍

skate-plugin/build.gradle.kts
kateliu20: Does this mean moving everything with aibot into a separate project? And what do you mean by adding it to project generator too?

ZacSweers: yeah I meant not having compose code directly in the `skate-plugin` project sources, we've struggled to keep that build setup working in the past. The project generator (which is also compose UI) is in a separate subproject, I'm suggesting you can move it there

kateliu20: Oh okay, that makes sense. I can rename the project-gen subproject to skate-ui and then add the aibot src code into that instead of skate-plugin

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
kateliu20: I couldn't find anything when searching for a trash icon in this link you provided: https://intellij-icons.jetbrains.design/. Instead, I was wondering if we could do a trailing icon that has an "x" button within the chat field. Something like this (instead of clear chat):
Screenshot 2024-08-27 at 3 05 24 PM

ZacSweers: image


But also I don't think we need one in the text field? That makes sense for a mobile UI where people have a thumb handy and the keyboard is unwieldy, but on desktop it's not a usual pattern. I'd suggest the delete icon to clear the chat and put it next to the "Clear Chat" text

kateliu20: I thought it looked a bit off that the Send button had just text and the Clear Chat had an icon, so I added a "send" icon to the send button as well. This way it looks a bit more consistent. What do you think?
Screenshot 2024-08-27 at 4 32 18 PM

kateliu20: Yeah, maybe just entering to send is easier, but I know DevXP AI and ChatGPT both have send buttons too, so it might be more popular with chatbots? That's also why I wanted to include it since those had it. But also, we can definitely discuss more and adjust in a future PR.

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
ZacSweers: yeah I mean can we clear its internal value rather than assigning a new instance. But also see the above about whether or not we need this at all :)

skate-plugin/src/main/kotlin/com/slack/sgp/intellij/aibot/ChatWindowCompose.kt
ZacSweers: hm, ok let's do something in the middle

- Keep the send button but make it disabled unless there's text inputted
- Let's make sure multi-line text and return-to-send work (shift+enter to newline, otherwise enter sends immediately)
- Let's remove the clear text button, it looks like  you have it for the chat text but I would expect this to clear the entire chat above if anything

kateliu20: That makes sense, I was able to do the first and third bullet, but I also don't know if Text Field supports shift+enter:
https://github.com/JetBrains/jewel/blob/main/ui/src/main/kotlin/org/jetbrains/jewel/ui/component/TextField.kt

I think maybe I should use a Basic Text Field instead. I will keep you posted. 
Screenshot 2024-08-28 at 12 04 24 PM
Screenshot 2024-08-28 at 12 04 38 PM

ZacSweers: I think it should be doable with `Modifier.onKeyEvent`.

Last bit - is the send text redundant here? The other examples I see are icon button only (with a content description)

kateliu20: Yeah good point, I can delete the send and add a content description instead. Thanks for the tip!

kateliu20: How is this for shift+enter and enter so far?
https://github.com/user-attachments/assets/fbd89bfc-aa76-4c36-b648-9475d4fe4bbd

ZacSweers: nice!

kateliu20:
It seems like after trying to migrate to Intellij Platform Gradle Plugin 2.x, the ./gradlew check fails because 

`LightJavaCodeInsightFixtureTestCase` doesn't seem to exist anymore. I'm not sure if it's because of the new java version, or if we need to [set it up](https://plugins.jetbrains.com/docs/intellij/testing-faq.html#how-to-test-a-jvm-language), but I think that there was probably some issue in trying to migrate this that it failed. I tried different imports [shown here ](https://plugins.jetbrains.com/docs/intellij/light-and-heavy-tests.html#-f141ga_24) but nothing related to java can import in this case.

PR #937, opened 2024-08-28
Update jewel to 0.23.0

ZacSweers:
The latest stable Android Studio is now based on intellij 2024.1, so we can move up to a newer Jewel version. This also moves Skate to use the latest compose-multiplatform alphas (but still only compiles SGP against the latest stable gradle plugin).


kateliu20 (approved):
This is good to know, thanks!

PR #948, opened 2024-09-05
Skate AI Initial Chat User Text Field and Window

kateliu20:
This is a continuation of #927, but because of the Jewel update in #937, I ended up running into some hiccups with transitioning the project and migrating to a different version of intellij gradle platform plugin. We reverted it back in #947 and this PR reflects the changes that are compatible with the old Jewel version (which is back on main). 

I have included the functionality for user input, where send button clears text, enter clears text, and shift + enter adds new lines. Users can also add multiple lines. I also added a function where the send button is faded and unselectable until there is text input. In future PRs, I will add the chat conversation as well, and this is just a start.

https://github.com/user-attachments/assets/94971a43-476e-4d67-924c-e012c633bc3c

### DIFF
--- a/skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
+++ b/skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
@@ -115,5 +115,5 @@ fun ConversationField(modifier: Modifier = Modifier) {
 }
 
 fun Modifier.enabled(enabled: Boolean): Modifier {
-  return this.then(if (enabled) Modifier else Modifier.alpha(0.5f))
+  return this.then(if (enabled) Modifier.alpha(1.0f) else Modifier.alpha(0.38f))
 }
--- a/skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
+++ b/skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
@@ -61,6 +61,8 @@ fun ConversationField(modifier: Modifier = Modifier) {
     horizontalArrangement = Arrangement.Center,
     verticalAlignment = Alignment.Bottom,
   ) {
+    // user text input. enter sends the message and clears text
+    // shift + enter adds a new line and expands the text field
     TextArea(
       value = textValue,
       onValueChange = { newText -> textValue = newText },
@@ -84,7 +86,6 @@ fun ConversationField(modifier: Modifier = Modifier) {
                 true
               }
             }
-
             else -> false
           }
         },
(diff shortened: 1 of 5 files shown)


skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
```diff
@@ -0,0 +1,133 @@
...
+    ConversationField(modifier)
+  }
+}
+
+@Composable
+fun ConversationField(modifier: Modifier = Modifier) {
+  var textValue by remember { mutableStateOf(TextFieldValue()) }
+  val isTextNotEmpty = textValue.text.isNotBlank()
+  Row(
+    modifier.fillMaxWidth().height(IntrinsicSize.Min).padding(4.dp),
```
ZacSweers, line 70: What's the height() modifier doing in this case?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
```diff
@@ -0,0 +1,133 @@
...
+              modifier = Modifier.size(20.dp),
+            )
+          }
+        }
+      }
+    }
+  }
+}
+
+fun Modifier.fadeWhenDisabled(enabled: Boolean): Modifier {
```
ZacSweers, line 131: Fade is usually an animation, but this isn't in this case. Let's call it something like `enabled` and make it more about its semantic meaning.

Also, material has a pattern for this we could follow from its `ContentAlpha` class - https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material/material/src/commonMain/kotlin/androidx/compose/material/ContentAlpha.kt;l=58?q=ContentAlpha

ZacSweers (approved):
A few comments before you merge. Also - let's do a small pass to add some clarifying comments in some of the meatier parts of the code, like the text handling

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
kateliu20: I used it to constrain the height of the row because IntrinsincSize.Min calculates the minimum height of the child which would be the text field.

ZacSweers: is that not what it does automatically?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
kateliu20: Ah okay, yeah this makes sense! I will add this adjustment to the next PR then.

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
kateliu20: It also helps with the height with the icon button too I think. I followed this https://medium.com/@oleg7green/intrinsic-measurements-in-jetpack-compose-f96e0d34e9e9. Without this, the send button doesn't appear or isn't aligned correctly.

ZacSweers: Can you screenshot what you mean? Having trouble picturing it

kateliu20: Screenshot 2024-09-05 at 1 16 42 PM
Like this.

PR #950, opened 2024-09-05
Chat Bubble and "Conversation" 

kateliu20:
In this PR, I added a Chat bubble that will print back the user input. I also set up a function "call api" that will call the DevXP API in the future, but currently sends back the user input as well. The conversation field is also scrollable (with no scroll bar at the moment). 

I am also worried about how there is not much padding once the text input becomes multiple lines (like in example 2 of the video), and was wondering if I could get advice on how to add it, because adding padding modifiers seems to only change the box and not the inner text, and the height parameter is what creates the initial space in the text field.

https://github.com/user-attachments/assets/50c6bf01-a77a-4b4d-950e-a849ad70ef16

EDIT: here is an update based on feedback. Now, the user's messages are on the right and response is on the left.

https://github.com/user-attachments/assets/62170207-3697-41e5-ac0e-3ff24b4cb1fd

### DIFF
--- a/skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
+++ b/skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
@@ -28,6 +28,7 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.wrapContentWidth
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.shape.RoundedCornerShape
@@ -64,7 +65,14 @@ fun ChatWindow(modifier: Modifier = Modifier) {
   var messages by remember { mutableStateOf(listOf<Message>()) }
   Column(modifier = Modifier.fillMaxSize().background(JewelTheme.globalColors.paneBackground)) {
     LazyColumn(modifier = Modifier.weight(1f), reverseLayout = true) {
-      items(messages.reversed()) { message -> ChatBubble(message) }
+      items(messages.reversed()) { message ->
+        Row(
+          modifier = Modifier.fillMaxWidth(),
+          horizontalArrangement = if (message.isMe) Arrangement.End else Arrangement.Start,
+        ) {
+          ChatBubble(message)
+        }
+      }
     }
     ConversationField(
       modifier = modifier,
@@ -150,7 +158,7 @@ fun ConversationField(modifier: Modifier = Modifier, onSendMessage: (String) -> 
 @Composable
 fun ChatBubble(message: Message, modifier: Modifier = Modifier) {
   Box(
-    Modifier.fillMaxWidth()
+    Modifier.wrapContentWidth()
       .padding(8.dp)
       .shadow(elevation = 0.5.dp, shape = RoundedCornerShape(25.dp), clip = true)
       .background(
--- a/skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
+++ b/skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
@@ -16,7 +16,18 @@
 package slack.tooling.aibot
 
 import androidx.compose.foundation.background
-import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.IntrinsicSize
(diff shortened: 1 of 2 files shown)


skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
```diff
@@ -34,30 +39,44 @@ import androidx.compose.ui.input.key.onPreviewKeyEvent
...
 import androidx.compose.ui.text.TextRange
+import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.input.TextFieldValue
 import androidx.compose.ui.unit.dp
 import org.jetbrains.jewel.foundation.theme.JewelTheme
-import org.jetbrains.jewel.ui.component.Icon
-import org.jetbrains.jewel.ui.component.IconButton
-import org.jetbrains.jewel.ui.component.Text
-import org.jetbrains.jewel.ui.component.TextArea
+import org.jetbrains.jewel.ui.component.*
```
ZacSweers, line 46: Can you fix these star imports in your IDE settings?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
```diff
@@ -34,30 +39,44 @@ import androidx.compose.ui.input.key.onPreviewKeyEvent
...
 
 @Composable
 fun ChatWindow(modifier: Modifier = Modifier) {
```
ZacSweers, line 49: We're getting a little to the point now logic-wise that I think it would be wise to start using Circuit for this. Doesn't have to be in this PR, but I think it's probably worth it for the next PR

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
```diff
@@ -114,6 +133,41 @@ fun ConversationField(modifier: Modifier = Modifier) {
   }
 }
 
+@Composable
+fun ChatBubble(message: Message, modifier: Modifier = Modifier) {
+  Box(
+    Modifier.fillMaxWidth()
```
ZacSweers, line 139: Does this need to fill the max width? It looks a little awkward in your video

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
```diff
@@ -114,6 +133,41 @@ fun ConversationField(modifier: Modifier = Modifier) {
...
   return this.then(if (enabled) Modifier.alpha(1.0f) else Modifier.alpha(0.38f))
 }
+
+fun callApi(message: String): String {
+  // function set up to call the DevXP API in the future.
+  // right now, just sends back the user input message
+  return (message)
+}
+
+object ChatColors {
```
ZacSweers, line 166: nice 👍

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
kateliu20: You have a point, I can adjust to make it less wide.

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
ZacSweers: yes we don't like them :). Preferable to be explicit

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
kateliu20: Should I adjust it so that it's based on the input text? Here are some examples:
Screenshot 2024-09-05 at 4 49 41 PM

ZacSweers: yep that makes sense to me 👍

ZacSweers: though seems we should put the user's messages on one side and the bot's messages on the other?

kateliu20: I put the updated video in the main description.

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindow.kt
ZacSweers: nit before you merge: Let's pull this to its own file as I suspect you'll keep adding to it. Would be great if we could pull these from the Jewel theme too

ZacSweers (approved):

PR #955, opened 2024-09-10
Add Circuit logic to AI Bot

kateliu20:
Adding Circuit logic, which has a Chat Screen state, UI (ChatWindowUi object) and Presenter which handles the sending logic. I followed the [tutorial](https://slackhq.github.io/circuit/tutorial/#__tabbed_2_1) and Project Gen for reference, but I also haven't used Circuit before, so I would appreciate any feedback. 

f27ddc1af95d1c0b90e33622407588cd464def09


skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatColors.kt
```diff
@@ -15,21 +15,18 @@
...
-import org.jetbrains.jewel.foundation.theme.JewelTheme
 
 object ChatColors {
   val promptBackground = Color(0xFF45494A)
 
   // Color(0xFF2d2f30) responseBackground
-  val responseBackground: Color
-    @Composable @ReadOnlyComposable get() = JewelTheme.globalColors.infoContent
+  val responseBackground = Color(0xFF2d2f30)
+  //    @Composable @ReadOnlyComposable get() = JewelTheme.globalColors.infoContent
```
ZacSweers, line 25: ?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatPanel.kt
```diff
@@ -15,16 +15,31 @@
...
     }
   }
+
+  @Composable
+  private fun ChatApp() {
+    val circuit =
+      Circuit.Builder()
+        .addPresenter<ChatScreen, ChatScreen.State>(ChatPresenter())
+        .addUi<ChatScreen, ChatScreen.State> { state, modifier -> ChatWindow(state, modifier) }
+        .build()
```
ZacSweers, line 41: you should wrap this in a `remember { }` block, otherwise it'll get recreated every time

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatPresenter.kt
```diff
@@ -0,0 +1,52 @@
...
+    var currentInput by remember { mutableStateOf("") }
+
+    return ChatScreen.State(messages = messages, currentInput = currentInput) { event ->
+      when (event) {
+        is ChatScreen.Event.SendMessage -> {
+          val newMessage = Message(event.message, isMe = true)
+          messages = messages + newMessage
+          val response = Message(callApi(event.message), isMe = false)
+          messages = messages + response
+          currentInput = "" // clear chat after sending
```
ZacSweers, line 38: I know this is just stubbed out, but for future ref you don't wanna have too much logic in this even callback directly. Instead, you generally want it to just update internal state.

ZacSweers, line 38: I would suggest keeping "currentInput" entirely in the UI, and just use this event to signal a new sent message with the text received from the UI. What do you think?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatScreen.kt
```diff
@@ -0,0 +1,34 @@
...
+  data class State(
+    val messages: List<Message>,
+    val currentInput: String,
+    val eventSink: (Event) -> Unit,
+  ) : CircuitUiState
+
+  sealed class Event : CircuitUiEvent {
+    data class SendMessage(val message: String) : Event()
+
+    data class UpdateInput(val input: String) : Event()
```
ZacSweers, line 32: What does this do?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatScreen.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import com.slack.circuit.runtime.CircuitUiEvent
+import com.slack.circuit.runtime.CircuitUiState
+import com.slack.circuit.runtime.screen.Screen
+
+data object ChatScreen : Screen {
+  data class State(
+    val messages: List<Message>,
+    val currentInput: String,
+    val eventSink: (Event) -> Unit,
```
ZacSweers, line 26: nit: for testability, we usually default event sinks to `= {}` for ease of standing up new state objects

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindowUi.kt
```diff
@@ -0,0 +1,178 @@
...
+import androidx.compose.ui.unit.dp
+import org.jetbrains.jewel.foundation.theme.JewelTheme
+import org.jetbrains.jewel.ui.component.Icon
+import org.jetbrains.jewel.ui.component.IconButton
+import org.jetbrains.jewel.ui.component.Text
+import org.jetbrains.jewel.ui.component.TextArea
+
+object ChatWindowUi {
+  @Composable
+  fun ChatWindow(state: ChatScreen.State, modifier: Modifier = Modifier) {
```
ZacSweers, line 64: Generally we just keep these functions top-level, no need for the wrapper object

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindowUi.kt
```diff
@@ -0,0 +1,178 @@
...
+        modifier = Modifier,
+        onSendMessage = { userMessage ->
+          state.eventSink(ChatScreen.Event.SendMessage(userMessage))
+        },
+      )
+    }
+  }
+
+  @Composable
+  fun ConversationField(modifier: Modifier = Modifier, onSendMessage: (String) -> Unit) {
```
ZacSweers, line 86: nit: private should be the default

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/Message.kt
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package slack.tooling.aibot
+
+import androidx.compose.runtime.Immutable
+
+// not sure if I need this in a separate class
```
ZacSweers, line 20: Could you expand on this?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindowUi.kt
```diff
@@ -0,0 +1,178 @@
...
+                  sendMessage()
+                  true
+                }
+              }
+              else -> false
+            }
+          },
+        placeholder = { Text("Start your conversation") },
+        keyboardActions = KeyboardActions.Default,
+        maxLines = Int.MAX_VALUE,
```
ZacSweers, line 132: Did you mean to leave these defaults in?

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindowUi.kt
```diff
@@ -0,0 +1,178 @@
...
+            if (isTextNotEmpty) {
+              sendMessage()
+            }
+          },
+          enabled = isTextNotEmpty,
+        ) {
+          Icon(
+            painter = painterResource("/drawable/send.svg"),
+            contentDescription = "Send",
+            modifier = Modifier.size(20.dp),
```
ZacSweers, line 148: For this I would recommend using `Modifier.minimumInteractiveComponentSize()`, which is more dynamic and will respect user accessibility settings too

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatColors.kt
kateliu20: I reverted the chat colors back to the original because I didn't like how the Jewel Theme muted the colors 

This is the Jewel color:
Screenshot 2024-09-10 at 1 40 34 PM

This is the original:
Screenshot 2024-09-10 at 1 41 01 PM

I just commented out the logic in case we wanted to bring it back but I can delete these lines altogether.

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/Message.kt
kateliu20: Do you think it's okay to have the Message data class in a separate file? Or is it unnecessary?

ZacSweers: if you're going to reuse it yeah, otherwise you could just co-locate it with the rest of your state models

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindowUi.kt
kateliu20: I did initially because I was trying different things for the "shift+enter" functionality. I just realized that it doesn't make a difference, so I will remove it. Thanks for pointing it out!

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindowUi.kt
ZacSweers: Good point. Maybe look at their docs and see if they have anything comparable?

kateliu20: Could I use `wrapContentSize()` and set `unbounded = true` to account for accessibility? 

According to the docs: 
"Allow the content to measure at its desired size without regard for the incoming measurement minimum width or minimum height constraints, and, if unbounded is true, also without regard for the incoming maximum constraints. If the content's measured size is smaller than the minimum size constraint, align it within that minimum sized space. If the content's measured size is larger than the maximum size constraint (only possible when unbounded is true), align within the maximum space."

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatColors.kt
ZacSweers: Let's save that for another PR since it's sort of orthogonal to this one

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatPresenter.kt
kateliu20: Yeah this makes sense, I should just keep it in UI. Will make this change! Thanks

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatScreen.kt
kateliu20: I thought since `SendMessage` is an event, then I should be using `CircuitUiEvent` with a sealed class. That's how I set it up for the `ChatPresenter`. But after I removed `currentInput` from the presenter, I don't need that `UpdateInput` class.

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatPanel.kt
kateliu20: 2d42f95f8a9a4c4320581950aabb065bc1516b8d in ChatPanel.kt

kateliu20: f27ddc1af95d1c0b90e33622407588cd464def09 under `ChatPanel.kt`

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatPresenter.kt
kateliu20: I removed it in f27ddc1af95d1c0b90e33622407588cd464def09 under `ChatPresenter.kt`

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatWindowUi.kt
kateliu20: f27ddc1af95d1c0b90e33622407588cd464def09 in `ChatWindowUi.kt`

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatScreen.kt
kateliu20: I set it here: https://github.com/slackhq/slack-gradle-plugin/blob/f27ddc1af95d1c0b90e33622407588cd464def09/skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/ChatScreen.kt

with this commit: f27ddc1af95d1c0b90e33622407588cd464def09

I will make sure to separate my commits, sorry for the headache.

ZacSweers (approved):

PR #976, opened 2024-09-23
Update Jewel to Latest Version (0.24.2)

kateliu20:
Referencing https://github.com/slackhq/slack-gradle-plugin/pull/937 which was reverted due to not upgrading to IJ platform 2.0.1 yet, but now is done here: https://github.com/slackhq/slack-gradle-plugin/pull/973.


gradle/libs.versions.toml
```diff
@@ -4,7 +4,8 @@ agpAlpha = "8.6.1"
 anvil = "2.5.0-beta11"
 bugsnagGradle = "8.1.0"
 circuit = "0.23.1"
-compose-jb = "1.6.11"
+compose-jb = "1.7.0-alpha03"
```
ZacSweers, line 7: Let's use the latest beta - https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.7.0-beta02

gradle/libs.versions.toml
```diff
@@ -13,7 +14,7 @@ errorproneGradle = "3.0.1"
 intellij-platform = "2.0.1"
 jdk = "22"
 jvmTarget = "17"
-jewel = "0.15.2.2"
+jewel = "0.23.0"
```
ZacSweers, line 17: Let's use the latest version

gradle/libs.versions.toml
kateliu20: aa32432f1ca9f371e9c8ebfe96d96a369f54c217

gradle/libs.versions.toml
kateliu20: eae65724045d598a94a4c65b6fa00ea99ef68d22

Because of these last two commits, the painter resource for my Icon was deprecated, so I need to work on fixing that:
'painterResource(String): Painter' is deprecated. Migrate to the Compose resources library. See https:// www. jetbrains. com/ help/ kotlin-multiplatform-dev/ compose-images-resources. html If you need to load resources specificly from Java classpath, you should still use the new resource library and use a snippet from https:// github. com/ JetBrains/ compose-multiplatform-core/ pull/ 1457

kateliu20: I paired with Linh and updated it and now it works. I used the suggested migration snippet they posted https://github.com/JetBrains/compose-multiplatform-core/pull/1457 for the painter resource. I will keep an eye out in case there are any changes.

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/PainterResource.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.compose.ui.graphics.vector.rememberVectorPainter
+import androidx.compose.ui.platform.LocalDensity
+import org.jetbrains.compose.resources.ExperimentalResourceApi
+import org.jetbrains.compose.resources.decodeToImageBitmap
+import org.jetbrains.compose.resources.decodeToImageVector
+import org.jetbrains.compose.resources.decodeToSvgPainter
+
+// Migration snippet copied from https://github.com/JetBrains/compose-multiplatform-core/pull/1457
+// To resolve deprecated painterResource function from upgrading compose-jb
+class PainterResource {
```
ZacSweers, line 31: This should be an `object`, otherwise you're gonna create a new instance every time. In general - if the class has no internal state, you should default to using top-level functions instead. Maaaaybe an `object` class if you really need the namespacing.

kateliu20: Thanks for the tip, I changed it here: 35b22fcc01010d9856c5c824f73034297330f157

skate-plugin/project-gen/src/jvmMain/kotlin/slack/tooling/aibot/PainterResource.kt
```diff
@@ -28,7 +28,7 @@ import org.jetbrains.compose.resources.decodeToSvgPainter
 
 // Migration snippet copied from https://github.com/JetBrains/compose-multiplatform-core/pull/1457
 // To resolve deprecated painterResource function from upgrading compose-jb
-class PainterResource {
+object PainterResource {
```
ZacSweers, line 31: Does it need to be an object vs top-level functions?

kateliu20: Ah yeah it doesn't need to be top-level. I wrapped it in the class originally because I was afraid it would call the deprecated painterResource function, but since I removed the import it should be good. I think I should remove the object :
d432093d8eb493b111c741a093ee93b0bdd59d71

ZacSweers (approved):
Before you merge let's do a quick sanity test running the plugin in studio and runIde works, just since we got burned before 👍

kateliu20:
> Before you merge let's do a quick sanity test running the plugin in studio and runIde works, just since we got burned before 👍

Yeah I checked that already! Definitely didn't want a repeat. I will continue to make adjustments to the UI independently and after talking with designers.
Screenshot 2024-09-23 at 5 48 34 PM

PR #1004, opened 2024-10-02
Add AIBot to Settings (Enable/Disable)

kateliu20:
I added it to the settings so users can disable the chatbot if they don't want to see it. This is also good just to have in the settings so we can add more (like calling the API) to this. 

I changed it from a tool window to a tool window listener. Here's a demo below:

https://github.com/user-attachments/assets/9c9c9faa-e31d-4079-a915-caf356e6a403

I also wasn't sure if I'm still able to include the lightning icon. If there are any suggestions on how to do that please let me know! Thank you!


ZacSweers:
> I added it to the settings so users can disable the chatbot if they don't want to see it. 

What's the difference between this and just closing the toolwindow? Is the bot running actively in the background otherwise?

PR #1018, opened 2024-10-09
Add isLoading State to Chat

kateliu20:
Adding a state for if it's loading once the API is called, so that a loading animation can be added while waiting for the chat response.

Below is an example of the animation. I enabled it so it's always loading just to show the animation (since the API call hasn't been merged yet)
https://github.com/user-attachments/assets/7abb3041-7aa0-4117-94e6-710a6b0cb800


platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/ChatWindowUi.kt
```diff
@@ -70,15 +70,23 @@ fun ChatWindowUi(state: ChatScreen.State, modifier: Modifier = Modifier) {
         }
       }
     }
+    if (state.isLoading) {
+      //       insert loading animation
```
ZacSweers, line 74: It looked like you had this done in your demo already, want to just add it in this PR?

kateliu20: 750f1ac0a65e1999a490405b671f03207cc00230

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
```diff
@@ -0,0 +1,87 @@
...
+
+@Composable
+fun LoadingAnimation(
+  modifier: Modifier = Modifier,
+  dotSize: Dp = 7.dp,
+  dotColor: Color = JewelTheme.contentColor,
+  spacing: Dp = 7.dp,
+  movementDistance: Dp = 10.dp,
+) {
+  val animatedDots = List(4) { remember { Animatable(0f) } }
```
ZacSweers: This will re-allocated every time, shouldn't the remember be outside of the `List` creation?

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/ChatPresenter.kt
```diff
@@ -36,7 +36,7 @@ class ChatPresenter : Presenter<ChatScreen.State> {
           isLoading = true
           val response = Message(callApi(event.message), isMe = false)
           messages = messages + response
-          isLoading = false
+          //          isLoading = false commented out to test loading animation
```
ZacSweers: Let's not leave this in when merging

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
kateliu20: Yeah, you are right. The remember should be outside the List block because otherwise it would be calling remember on each item of the list. Thank you for catching that, I made the change here: 3ca30396d163bb823843057e0ba62aea5a66f24c

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/ChatPresenter.kt
kateliu20: Thanks, fixed here: 5f8baf949662c19f7c1ae23ceba5b5559e9de40f

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
```diff
@@ -0,0 +1,87 @@
...
+import androidx.compose.ui.graphics.graphicsLayer
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import kotlinx.coroutines.delay
+import org.jetbrains.jewel.foundation.theme.JewelTheme
+
+// tutorial from https://www.youtube.com/watch?v=xakNOVaYLAg
+
+@Composable
```
ZacSweers, line 43: nit: let's make this a doc on the function rather than this floating comment, that way it's attached in kdoc

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
```diff
@@ -0,0 +1,87 @@
...
+                0.0f at 600 using LinearOutSlowInEasing
+                0.0f at 1200 using LinearOutSlowInEasing
+              },
+            repeatMode = RepeatMode.Restart,
+          ),
+      )
+    }
+  }
+
+  val animatedValues = animatedDots.map { it.value }
```
ZacSweers, line 74: This will reallocate every time too. Do you maybe want a SnapshotStateList instead?

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/ChatWindowUi.kt
```diff
@@ -131,13 +139,13 @@ private fun ConversationField(modifier: Modifier = Modifier, onSendMessage: (Str
     Column(Modifier.fillMaxHeight(), verticalArrangement = Arrangement.Center) {
       // button will be disabled if there is no text
       IconButton(
-        modifier = Modifier.padding(4.dp).enabled(isTextNotEmpty),
+        modifier = Modifier.padding(4.dp).enabled(isTextNotEmpty && !isLoading),
```
ZacSweers, line 142: `IconButton` has an `enabled` parameter we can use, any reason not to use that? That should handle internal state better too

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
kateliu20: Is this okay? 843a85b025b0f564952bcdfd208f11ef2b1bc8bc

ZacSweers: Yeah but let's add a proper description to it too, make the source just a footnote at the end for reference

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/ChatWindowUi.kt
kateliu20: I did it mainly for how it looks. The enabled parameter doesn't make the button less faded, so I wanted to make it more obvious it's disabled when I set it.

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
kateliu20: Sounds good, so does It prevent the recreation of our Animatable objects on each recomposition? I implemented a change here: 23bf68de2e6691756662189c46170cd1210e3b4a

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
kateliu20: Does this description look good? 23bf68de2e6691756662189c46170cd1210e3b4a

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/ChatWindowUi.kt
ZacSweers: interesting. Ok let's keep this but file an issue on jewel about making the button disabled appearance reflect it, feels like a bug that it doesn't

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
ZacSweers: Not quite. The kdoc should say what the composable itself does, not just where it's from. Let's do both

```kotlin
/**
 * This shows a loading indicator of three bouncing dots, etc etc.
 *
 * @see <a href="https://...">Source</a>
 */
```

kateliu20: Sounds good, thanks! d6d709817002cbbd112499aa1a6a1ea5d6bb73ba 
I made the adjustment here.

kateliu20: Adjusted here: b457fec214ebbdbfcb43f01ab5de439336f40b6c

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
```diff
@@ -0,0 +1,99 @@
...
+import kotlinx.coroutines.delay
+import org.jetbrains.jewel.foundation.theme.JewelTheme
+
+/**
+ * This shows a loading animation of three bouncing dots that occurs when the user is waiting on a
+ * response.
+ *
+ * Adopted from the Three-Dot Loading Animation Tutorial with Jetpack Compose by Stevdza-San from
+ *
+ * @see <a href="https://www.youtube.com/watch?v=xakNOVaYLAg">Source</a>
```
ZacSweers, line 49: Nit before you merge: you can just use the `@see` tag, no need for both

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
ZacSweers: Actually - do we need this intermediate variable at all? i.e. why not just do `animatedDots.forEach { dot ->` in the below block?

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
```diff
@@ -0,0 +1,99 @@
...
+ */
+@Composable
+fun LoadingAnimation(
+  modifier: Modifier = Modifier,
+  dotSize: Dp = 7.dp,
+  dotColor: Color = JewelTheme.contentColor,
+  spacing: Dp = 7.dp,
+  movementDistance: Dp = 10.dp,
+) {
+
```
ZacSweers, line 59: nit: don't need this newline

ZacSweers (approved):
One last question, also link the issue you file on the jewel repo for posterity 👍

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/aibot/LoadingAnimation.kt
kateliu20: Oh yeah that's a great point. I can use dot.value instead too

PR #1026, opened 2024-10-10
Source blockmapFile from generator task

ZacSweers:


kateliu20 (approved):
What exactly is this new change doing? Are we setting a value to the blockmapFile here?

ZacSweers:
Yes, we have two tasks here. One generates the file and the other uses it. Before they weren't properly linked, now this plumbs the generated file (an output property) _into_ the second task's input property

PR #1064, opened 2024-10-28
Roborazzi Test for Markdown Playground

kateliu20:
I've added a Roborazzi test for Compose Desktop support (based on the suggestions [here](https://github.com/takahirom/roborazzi?tab=readme-ov-file#experimental-feature-compose-desktop-support)). I also added the necessary dependencies for testing and compose.


platforms/intellij/compose/playground/src/jvmMain/kotlin/foundry/intellij/compose/playground/MarkdownPlayground.kt
```diff
@@ -31,11 +33,20 @@ import org.jetbrains.jewel.intui.standalone.theme.IntUiTheme
 import org.jetbrains.jewel.ui.component.DefaultButton
 import org.jetbrains.jewel.ui.component.Text
 
-fun main() = singleWindowApplication {
+fun main() {
+  singleWindowApplication { AppContent() }
+}
+
+// created function for testing
```
kateliu20, line 40: I added a comment here for now since testing on the function wrapped in a singleWindowApplication seemed to result in an infinite loop (where the test wouldn't end). I made a composable `AppContent()` function with the same content and added a test tag too.

platforms/intellij/compose/playground/src/jvmTest/kotlin/foundry/intellij/compose/playground/MarkdownPlaygroundKtTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.onRoot
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.runDesktopComposeUiTest
+import com.github.takahirom.roborazzi.RoborazziOptions
+import io.github.takahirom.roborazzi.captureRoboImage
+import kotlinx.coroutines.test.runTest
+import org.junit.Test
+
+class MarkdownPlaygroundKtTest {
```
kateliu20, line 28: Here is the test, based on the sample one: https://github.com/takahirom/roborazzi?tab=readme-ov-file#experimental-feature-compose-desktop-support

platforms/intellij/compose/playground/build.gradle.kts
```diff
@@ -47,6 +49,14 @@ kotlin {
         implementation(projects.platforms.intellij.compose)
       }
     }
+    jvmTest {
+      dependencies {
+        implementation(libs.junit)
+        implementation(libs.roborazzi)
+        implementation(compose.desktop.common)
```
kateliu20, line 56: I added compose dependencies to the jvmTest. Before I added these dependencies, I had a `java.lang.NoClassDefFoundError`.

gradle/libs.versions.toml
```diff
@@ -147,4 +151,5 @@ slf4jNop = "org.slf4j:slf4j-nop:2.0.16"
 tikxml-htmlEscape = { module = "com.tickaroo.tikxml:converter-htmlescape", version = "0.8.15" }
 truth = "com.google.truth:truth:1.4.4"
 xmlutil-core = { module = "io.github.pdvrieze.xmlutil:core", version.ref = "xmlutil" }
-xmlutil-serialization = { module = "io.github.pdvrieze.xmlutil:serialization", version.ref = "xmlutil" }
\ No newline at end of file
+xmlutil-serialization = { module = "io.github.pdvrieze.xmlutil:serialization", version.ref = "xmlutil" }
+junit-junit = { group = "junit", name = "junit", version.ref = "junit" }
```
ZacSweers, line 155: We already have this dependency above :)

platforms/intellij/compose/playground/build.gradle.kts
```diff
@@ -56,3 +66,7 @@ configurations
   .configureEach { attributes { attribute(KotlinPlatformType.attribute, KotlinPlatformType.jvm) } }
 
 dependencies { lintChecks(libs.composeLints) }
+
+tasks.withType<KotlinCompile>().configureEach {
+  compilerOptions { freeCompilerArgs.add("-Xcontext-receivers") }
```
ZacSweers, line 71: Let's add a comment explaining why this is needed, and let's limit it to test JVM compilations.

```kotlin
tasks.named { it.name == "compileJvmTestKotlin" }...
```

Should be soooomething like that

platforms/intellij/compose/playground/build.gradle.kts
```diff
@@ -47,6 +49,14 @@ kotlin {
         implementation(projects.platforms.intellij.compose)
       }
     }
+    jvmTest {
+      dependencies {
+        implementation(libs.junit)
+        implementation(libs.roborazzi)
+        implementation(compose.desktop.common)
+        implementation(compose.desktop.uiTestJUnit4)
```
kateliu20, line 57: Would this also be the correct compose desktop test dependency to add?

ZacSweers (commented):
I would expect some screenshots to be checked in, are you going to add them? We should also enable git-lfs and enforce snapshots use it, you can see some example setup here: https://github.com/slackhq/circuit/pull/204

platforms/intellij/compose/playground/build.gradle.kts
ZacSweers: What does the sample in roborazzi use? Probably easiest to just match that to start

platforms/intellij/compose/playground/src/jvmTest/kotlin/foundry/intellij/compose/playground/MarkdownPlaygroundKtTest.kt
kateliu20: I used Intellij's "generate test" feature so it would place it in the right directory. It auto-generated the name. I can rename it to just `MarkdownPlaygroundTest`. Screenshot 2024-10-28 at 12 08 59 PM

platforms/intellij/compose/playground/build.gradle.kts
kateliu20: It seems [here](https://github.com/takahirom/roborazzi/blob/main/roborazzi-compose-desktop/build.gradle) they don't have dependencies in `jvmTest`, but they have these in the `commonMain` dependencies:

```
implementation(compose.runtime)
implementation(compose.desktop.currentOs)
implementation(project(":roborazzi-painter"))
implementation(libs.kotlin.stdlib.jdk8)
api libs.compose.ui.test.junit4.desktop
implementation libs.compose.ui.graphics.desktop
```

platforms/intellij/compose/playground/src/jvmTest/kotlin/foundry/intellij/compose/playground/MarkdownPlaygroundTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import androidx.compose.ui.test.runDesktopComposeUiTest
+import com.github.takahirom.roborazzi.RoborazziOptions
+import io.github.takahirom.roborazzi.captureRoboImage
+import kotlinx.coroutines.test.runTest
+import org.junit.Test
+
+class MarkdownPlaygroundKtTest {
+  @OptIn(ExperimentalTestApi::class)
+  @Test
+  fun test() = runTest {
```
hellohj, line 31: Bonus point. This test will test a light mode by default. If you want to test a dark mode, you can also try `@Config(qualifiers = "+night")` in another test.

platforms/intellij/compose/playground/src/jvmTest/kotlin/foundry/intellij/compose/playground/MarkdownPlaygroundTest.kt
```diff
@@ -0,0 +1,48 @@
...
+    runDesktopComposeUiTest {
+      setContent { AppContent() }
+
+      val roborazziOptions =
```
hellohj, line 35: If you define Roborazzi options/rule globally like [this](https://github.com/takahirom/roborazzi/tree/1d5b51d55e4c2b3cfaca13cf6d38bebc0d1fae23?tab=readme-ov-file#image-comparator-custom-settings), it will be easy to add more tests and you don't need to pass `roborazziOptions` whenever you call `captureRoboImage()` API.

platforms/intellij/compose/playground/src/jvmTest/kotlin/foundry/intellij/compose/playground/MarkdownPlaygroundTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        RoborazziOptions(
+          recordOptions = RoborazziOptions.RecordOptions(resizeScale = 0.5),
+          compareOptions = RoborazziOptions.CompareOptions(changeThreshold = 0F),
```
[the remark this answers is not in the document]
kateliu20: Yeah it seems like it [here](https://github.com/takahirom/roborazzi/pull/183). I removed it: 60903cfba4573cb758410c4e6499205b75d6acdf. Thanks!

gradle/libs.versions.toml
kateliu20: 😅 removed it: 60903cfba4573cb758410c4e6499205b75d6acdf thanks!

platforms/intellij/compose/playground/src/jvmTest/kotlin/foundry/intellij/compose/playground/MarkdownPlaygroundTest.kt
kateliu20: Is `RoborazziRule` supposed to be provided by the Roborazzi plugin? It was asking me to create my own class for `RoborazziRule` when I tried to add it.

hellohj: yes, you need to add this. `testing-roborazzi-rules = { module = "io.github.takahirom.roborazzi:roborazzi-junit-rule", version.ref = "roborazzi" }`

hellohj: These are things we use in our slack-android-ng. you may need to see which additional ones are necessary.
```
testing-roborazzi = { module = "io.github.takahirom.roborazzi:roborazzi", version.ref = "roborazzi" }
testing-roborazzi-rules = { module = "io.github.takahirom.roborazzi:roborazzi-junit-rule", version.ref = "roborazzi" }
testing-roborazzi-compose = { module = "io.github.takahirom.roborazzi:roborazzi-compose", version.ref = "roborazzi" }
testing-roborazzi-core = { module = "io.github.takahirom.roborazzi:roborazzi-core", version.ref = "roborazzi" }
```

kateliu20:
> I would expect some screenshots to be checked in, are you going to add them? We should also enable git-lfs and enforce snapshots use it, you can see some example setup here: [slackhq/circuit#204](https://github.com/slackhq/circuit/pull/204)

I'm having trouble getting the screenshots to actually generate. The test is successfully running, but the `captureRoboImage()` calls aren't producing any image files on my end. I am also trying to call it to a directory but nothing gets produced. I feel like the build files are set up but am I missing something in the test to produce images?

platforms/intellij/compose/playground/build.gradle.kts
ZacSweers: That's the library itself, we should look at one of their samples. I was eyeing this: https://github.com/takahirom/roborazzi/blob/main/sample-compose-desktop-jvm/build.gradle.kts

platforms/intellij/compose/playground/build.gradle.kts
kateliu20: Do I need to register `compileJvmTestKotlin` task? It's not in the list of gradlew tasks.

ZacSweers: the name might be slightly different, you'll want to find the task name for the task that compiles test jvm sources

kateliu20: I believe it could be `jvmTestClasses` under the `./gradlew tasks`

`jvmTestClasses - Assembles outputs for compilation 'test' of target 'jvm'` and added the change here: fff72b9fa44861e4a35431a293989dfacb0903dc

ZacSweers: jvmTestClasses is not a type of `KotlinCompile` as far as I know. Do you see this actually being configured there? I looked and I'm think what you want is `compileTestKotlinJvm`, but if it works without adding this opt-in at all then we can omit it?

kateliu20: I found the task `compileTestKotlinJvm` while running `./gradlew tasks --all `, so I committed it here: e1ebfee38af50a0dc12f7af6b590b77d48504466. Is this the task you are looking for?

ZacSweers: yes, in general you are gonna be better off not relying on `./gradlew tasks` vs just running a task you know will run it and adding `--dry-run` to the end, which will just print all the tasks it's gonna run. You can also look at a build scan to see task types.

.gitattributes
```diff
@@ -1,3 +1,4 @@
 gradlew linguist-generated=true
 gradlew.bat linguist-generated=true
 config/bin/* filter=lfs diff=lfs merge=lfs -text
+**/snapshots/**/*.png filter=lfs diff=lfs merge=lfs -text
```
kateliu20, line 4: I added this line to .gitattributes similar to the circuit [PR](https://github.com/slackhq/circuit/pull/204) I referenced. I also modified the folder name to be snapshots to match.

platforms/intellij/compose/playground/src/jvmTest/kotlin/foundry/intellij/compose/playground/MarkdownPlaygroundTest.kt
kateliu20: Do you know what dependency I need for this Config? I wasn't able to run the Roborazzi test due to a missing dependency, even though I have `robolectric` and `androidx-testing-core`. Is this annotation mobile specific?

config/git/hooks/pre-receive.sh
```diff
@@ -0,0 +1,8 @@
+# compares files that match .gitattributes filter to those actually tracked by git-lfs
+diff <(git ls-files ':(attr:filter=lfs)' | sort) <(git lfs ls-files -n | sort) >/dev/null
```
kateliu20, line 2: I'm able to run this script, but I'm not sure if it's running the expected behavior. 
Screenshot 2024-10-30 at 5 34 47 PM

platforms/intellij/compose/playground/src/jvmMain/kotlin/foundry/intellij/compose/playground/MarkdownPlayground.kt
ZacSweers: I don't think we need the comment, feels obvious in source why it's there :)

platforms/intellij/compose/playground/build.gradle.kts
```diff
@@ -56,3 +76,12 @@ configurations
   .configureEach { attributes { attribute(KotlinPlatformType.attribute, KotlinPlatformType.jvm) } }
 
 dependencies { lintChecks(libs.composeLints) }
+
+// Roborazzi Desktop support uses Context Receivers
+// This setup is for JVM test builds only
+tasks.named<KotlinCompile>("compileTestKotlinJvm").configure {
+  compilerOptions {
+    incremental = false
```
ZacSweers, line 84: What's this for?

platforms/intellij/compose/playground/build.gradle.kts
```diff
@@ -14,12 +14,18 @@
...
 plugins {
   alias(libs.plugins.kotlin.multiplatform)
   alias(libs.plugins.compose)
   alias(libs.plugins.kotlin.plugin.compose)
   alias(libs.plugins.lint)
+  alias(libs.plugins.roborazzi)
+}
+
+roborazzi {
+  outputDir = file("src/jvmTest/kotlin/foundry/intellij/compose/playground/snapshots/images")
```
ZacSweers, line 28: Is there a property input of this? We should ideally use this in gradle for lazier eval

```kotlin
outputDir.set(layout.projectDirectory.dir("src/jvmTest/kotlin/foundry/intellij/compose/playground/snapshots/images"))
```

if `outputDir` is just a `File` and not a `Property<File>`, then you can still use the above but just write `.asFile.get()` at the end


```kotlin
outputDir = layout.projectDirectory.dir("src/jvmTest/kotlin/foundry/intellij/compose/playground/snapshots/images").asFile.get()
```

gradle/libs.versions.toml
```diff
@@ -139,11 +143,17 @@ okio-fakefilesystem = { module = "com.squareup.okio:okio-fakefilesystem", versio
 oshi = "com.github.oshi:oshi-core:6.6.5"
 retrofit = { module = "com.squareup.retrofit2:retrofit", version.ref = "retrofit" }
 retrofit-converters-wire = { module = "com.squareup.retrofit2:converter-wire", version.ref = "retrofit" }
+robolectric = { group = "org.robolectric", name = "robolectric", version.ref = "robolectric" }
```
ZacSweers, line 146: talked offline but we shouldn't need this

.github/workflows/ci.yml
ZacSweers, line 1: We'll need to add a validation task to the "Build and run tests" step in this workflow too

.github/workflows/ci.yml
```diff
@@ -55,6 +55,9 @@ jobs:
           path: |
             **/build/reports/**
 
+      - name: Check LFS files
```
ZacSweers, line 58: let's check this earlier, right after the checkout step. Earlier feedback the better

platforms/intellij/compose/playground/build.gradle.kts
kateliu20: https://github.com/slackhq/foundry/pull/1064#discussion_r1819340674

I was trying to specify just needing the task name for "compileTestKotlinJvm"

.github/workflows/ci.yml
kateliu20: Removed here: 7a13dfd78af8e67ca8b8cb0f1b4dae15a26c909f

.github/workflows/ci.yml
```diff
@@ -41,7 +44,7 @@ jobs:
       - name: Build and run tests
         id: gradle
         timeout-minutes: 10
-        run: ./gradlew check
+        run: ./gradlew check verifyRoborazzi
```
takahirom, line 47: This is just a personal observation, but I prefer using `verifyRoborazzi**Debug**` since the test runs twice - once for debug and once for release.

takahirom: I apologize, but since this might be a JVM module, there may not be any build types available. 🙇

.github/workflows/ci.yml
ZacSweers: Did you link the right sha? Those changes look unrelated 😅

.github/workflows/ci.yml
```diff
@@ -54,6 +57,7 @@ jobs:
           name: reports
           path: |
             **/build/reports/**
+            **/src/jvmTest/snapshots/**/*_compare.png
```
ZacSweers, line 60: 👍

.github/workflows/ci.yml
```diff
@@ -63,6 +67,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          lfs: 'true'
```
ZacSweers, line 71: I don't think we need this for the snapshot publishing

.github/workflows/ci.yml
```diff
@@ -23,6 +23,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
```
ZacSweers, line 24: We do need the lfs checkout for the main CI job here right?

platforms/intellij/compose/gradle.properties
```diff
@@ -2,3 +2,4 @@ POM_ARTIFACT_ID=skate-compose
 POM_NAME=Skate (Compose)
 POM_DESCRIPTION=Skate (Compose)
 INTELLIJ_PLUGIN=true
+roborazzi.record.filePathStrategy=relativePathFromRoborazziContextOutputDirectory
```
ZacSweers, line 5: Can we add a comment for posterity on this?

platforms/intellij/compose/playground/src/jvmMain/kotlin/foundry/intellij/compose/playground/MarkdownPlayground.kt
```diff
@@ -31,11 +33,19 @@ import org.jetbrains.jewel.intui.standalone.theme.IntUiTheme
 import org.jetbrains.jewel.ui.component.DefaultButton
 import org.jetbrains.jewel.ui.component.Text
 
-fun main() = singleWindowApplication {
+fun main() {
+  singleWindowApplication { MarkdownPlayground() }
+}
```
ZacSweers, line 38:
```suggestion
fun main() = singleWindowApplication { MarkdownPlayground() }
```

platforms/intellij/compose/playground/src/jvmMain/kotlin/foundry/intellij/compose/playground/MarkdownPlayground.kt
```diff
@@ -31,11 +33,19 @@ import org.jetbrains.jewel.intui.standalone.theme.IntUiTheme
...
+
+@Composable
+fun MarkdownPlayground(modifier: Modifier = Modifier) {
   var isDark by remember { mutableStateOf(false) }
   IntUiTheme(isDark) {
-    Column(Modifier.background(JewelTheme.globalColors.panelBackground)) {
-      DefaultButton(modifier = Modifier.padding(16.dp), onClick = { isDark = !isDark }) {
+    Column(modifier.background(JewelTheme.globalColors.panelBackground)) {
+      DefaultButton(
+        modifier = Modifier.padding(16.dp).testTag("dark-mode-toggle"),
```
ZacSweers, line 46: nit: pull the test tag out to a shared constant so you can reuse that reference in test code too. That way if you update its value both get the update. Usually it's something like a `object TestTags { ... }`

platforms/intellij/compose/playground/src/jvmTest/kotlin/foundry/intellij/compose/playground/MarkdownPlaygroundTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.github.takahirom.roborazzi.DEFAULT_ROBORAZZI_OUTPUT_DIR_PATH
+import com.github.takahirom.roborazzi.ExperimentalRoborazziApi
+import com.github.takahirom.roborazzi.RoborazziOptions
+import io.github.takahirom.roborazzi.captureRoboImage
+import kotlin.test.Test
+
+class MarkdownPlaygroundTest {
+  @OptIn(ExperimentalTestApi::class, ExperimentalRoborazziApi::class)
+  @Test
+  fun test() = runDesktopComposeUiTest {
```
ZacSweers, line 32: let's call it `snapshot()`

platforms/intellij/compose/playground/src/jvmTest/kotlin/foundry/intellij/compose/playground/MarkdownPlaygroundTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        recordOptions = RoborazziOptions.RecordOptions(resizeScale = 0.5),
+        compareOptions =
+          RoborazziOptions.CompareOptions(outputDirectoryPath = DEFAULT_ROBORAZZI_OUTPUT_DIR_PATH),
+      )
+
+    onRoot().captureRoboImage(roborazziOptions = roborazziOptions)
+
+    onNodeWithTag("dark-mode-toggle").performClick()
+
+    onRoot().captureRoboImage(roborazziOptions = roborazziOptions)
```
ZacSweers, line 46: slick

platforms/intellij/compose/playground/src/jvmTest/kotlin/foundry/intellij/compose/playground/snapshots/images/foundry.intellij.compose.playground.MarkdownPlaygroundTest.test_2_compare.png
ZacSweers, line 1: Did you intend to check in these two compare pngs? I thought that was only for diffing/debugging failures?

.github/workflows/ci.yml
kateliu20: Sorry, I meant to link this one: b7b18ff0007806228a4090d137a9041d4d185f83 😅

.github/workflows/ci.yml
kateliu20: Added the changes here for lfs checkout: 992b80dd7906e3e8b7bf586d7afed8d3bf7cfe69

platforms/intellij/compose/gradle.properties
kateliu20: Sounds good, I added a comment `# Set the default path for Roborazzi record images` 992b80dd7906e3e8b7bf586d7afed8d3bf7cfe69

platforms/intellij/compose/playground/src/jvmMain/kotlin/foundry/intellij/compose/playground/MarkdownPlayground.kt
kateliu20: I added this object and referred to it here and in the test: 992b80dd7906e3e8b7bf586d7afed8d3bf7cfe69

platforms/intellij/compose/playground/src/jvmTest/kotlin/foundry/intellij/compose/playground/MarkdownPlaygroundTest.kt
kateliu20: Got it, done here too: 992b80dd7906e3e8b7bf586d7afed8d3bf7cfe69

ZacSweers (approved):
in the future let's split [this](https://github.com/slackhq/foundry/commit/992b80dd7906e3e8b7bf586d7afed8d3bf7cfe69) up into multiple commits. Having to search in the commit for the bits referenced by a comment adds a bit of extra work to reviewers :slightly_smiling_face:

PR #1082, opened 2024-11-11
Paper over test fixture dependency cycles in mod score computation

ZacSweers:
See the comments in code for why we're allowing this. Hopefully just a stopgap until we can finally use native test fixtures everywhere.


kateliu20 (approved):

CamiloVega:
Thanks

PR #1111, opened 2024-11-21
Enable CC sharing on CI

ZacSweers:
From https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:secrets:configuring_encryption_key

and how androidx does it https://github.com/androidx/androidx/blob/a7f7cfb8087fbcfd117c11c188d69c9a741a17ed/.github/workflows/presubmit.yml#L41

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,8 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      GRADLE_ENCRYPTION_KEY: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


kateliu20:
Is there a particular reason why we need to encrypt the configuration cache now, or is this a new update that allows this to happen?

ZacSweers:
It was always encrypted before, the change now is that gradle exposed an API to specify the encryption key used. This means that we can share a key that allows us to reuse the cache entries across CI runs, whereas before they were always unique and thus not shareable!

kateliu20 (approved):
Looks good👍

kateliu20:
> It was always encrypted before, the change now is that gradle exposed an API to specify the encryption key used. This means that we can share a key that allows us to reuse the cache entries across CI runs, whereas before they were always unique and thus not shareable!

That's cool! Would this allow for faster builds then, if the runs can reuse the cache entries now?

ZacSweers:
yep exactly

PR #1135, opened 2024-12-10
Add compressAndroidTestApksWithLegacyPackaging option

ZacSweers:


ZacSweers:
The CI failure is a red herring, due to a broken config cache config that is fixed on main but polluted atm

kateliu20 (approved):

platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/FoundryProperties.kt
```diff
@@ -459,6 +459,26 @@ internal constructor(
...
+    get() = resolver.booleanProvider("foundry.android.test.compressWithLegacyPackaging", false)
+
+  /**
+   * Flag for minifying androidTest APks with R8. This just tree shakes.
+   *
+   * See:
+   * - https://issuetracker.google.com/issues/259832799
+   * - https://developer.android.com/reference/tools/gradle-api/7.1/com/android/build/api/dsl/DexPackagingOptions#useLegacyPackaging:kotlin.Boolean
+   */
+  public val minifyAndroidTestApks: Provider<Boolean>
```
valeraz, line 479: could we rename this to `compressAndroidTestApks`. I'm concerned that minify will get conflated with running shrinking.

ZacSweers: We already have a compress property above this, this is the toe-hold for minification with R8 if there ends up being a way to do that too

PR #1185, opened 2025-01-23
Add orchestrator dep when androidTest is enabled

valeraz:
This is needed in order to run instrumentation tests locally.

### DIFF
--- a/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/StandardProjectConfigurations.kt
+++ b/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/StandardProjectConfigurations.kt
@@ -510,7 +510,9 @@ internal class StandardProjectConfigurations(
                 .getOrElse(false)
           val isAndroidTestEnabled = variant is HasAndroidTest && variant.androidTest != null
           if (isAndroidTestEnabled) {
-            dependencies.add("androidTestUtil", "androidx.test:orchestrator")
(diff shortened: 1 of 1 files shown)


platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/StandardProjectConfigurations.kt
```diff
@@ -510,6 +510,7 @@ internal class StandardProjectConfigurations(
                 .getOrElse(false)
           val isAndroidTestEnabled = variant is HasAndroidTest && variant.androidTest != null
           if (isAndroidTestEnabled) {
+            dependencies.add("androidTestUtil", "androidx.test:orchestrator")
```
ZacSweers, line 513: Before you merge, we have a property for this that we should probably still gate this on just to be safe. `FoundryProperties.useOrchestrator`

ZacSweers (approved):

platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/StandardProjectConfigurations.kt
ZacSweers: I think we keep it on by default and have the killswitch in place since we already have the property available anyway, doesn't hurt to have around

ZacSweers (approved):
Approvals are sticky in this repo fyi :)

PR #1207, opened 2025-02-10
Add configuration for Roborazzi snapshot testing

hellohj:
- This PR centralizes Roborazzi snapshot testing configuration, eliminating redundant code in subproject Gradle files and simplifying maintenance.  Engineers will no longer need to modify these files for snapshot testing.
- A test category filter has also been implemented to isolate Roborazzi snapshot tests from other unit tests within subprojects. Currently, all unit tests run together. With this change, when `verify/recordRoborazzi` command is executed with `-Pfoundry.test.snapshot=true` and Roborazzi tests are marked with a category from `foundry.test.snapshot.category`, only the snapshot tests will be executed in CI.

- **TODO**: we need to figure out the optimal location for `project.pluginManager.apply("io.github.takahirom.roborazzi")`.  Initially, it was added to `FoundryExtension` alongside the Roborazzi snapshot changes, but this resulted in the following error: 
```
It is too late to add actions as the callbacks already executed.
Did you try to call beforeVariants or onVariants from the old variant API
'applicationVariants' for instance ? you should always call beforeVariants or
onVariants directly from the androidComponents DSL block.
```


ZacSweers:
@hellohj where did we try to apply the plugin before? I don't see it here 🤔

hellohj:
If I remember, we deleted the line `project.pluginManager.apply("io.github.takahirom.roborazzi")` from `applyTo()` because of the error from the PR description. When we moved this line to `snapshotTest()` method, it didn't recognize `project` so we couldn't do it. If you think there is a better place, please let me know. For my testing, I still included `alias(libs.plugins.roborazzi)` plugin to my ng gradle file.

ZacSweers (commented):
Found what I think is an ok solution and pushed it. Do you want to try this out again in your local clone and see how this works?

hellohj:
@ZacSweers I tested it and it works as expected. Thanks for the change!

ZacSweers (approved):

PR #1254, opened 2025-03-09
WIP update JVM target and IJ platform

ZacSweers:


ZacSweers:
Consolidating to #1254

PR #1255, opened 2025-03-10
Update JVM target to 21 and kotlin to 2.0 for gradle

ZacSweers:


.github/workflows/ci.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       - name: Build and run tests
         id: gradle
         timeout-minutes: 10
-        run: ./gradlew check
+        run: ./gradlew check --quiet
```
kateliu20, line 55: What's the `--quiet` flag for?

ZacSweers: It makes the CI only print errors when it fails, matches what we do elsewhere

kateliu20 (approved):

PR #1262, opened 2025-03-14
Add support for Java agent configuration in tests

ZacSweers:
Introduce a new `testJavaAgents` API for simplifying the configuration of Java agents in test environments. The basic entry point is

```kotlin
dependencies {
  testJavaAgents {
    javaAgent(...)
  }
}
```

With a convenience (deprecated) `mockitoAgent()` (which is what spurred this)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changelog
 **Unreleased**
 --------------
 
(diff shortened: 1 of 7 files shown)


kateliu20:
Is [this](https://github.com/mockito/mockito/issues/3037#issuecomment-2724222930) what you were referring to with the basic entry point?

ZacSweers:
Yeah, but we don't need a whole dedicated plugin as we really just want this for tests. That plugin is more for Java application projects

PR #1265, opened 2025-03-17
Use Gradle Problems API for module validation reporting

ZacSweers:
https://docs.gradle.org/current/userguide/reporting_problems.html. Not as rich as markdown but less noise and better integrated.

Before

image

After

image

### DIFF
--- a/platforms/gradle/foundry-gradle-plugin/build.gradle.kts
+++ b/platforms/gradle/foundry-gradle-plugin/build.gradle.kts
@@ -86,9 +86,6 @@ dependencies {
   implementation(libs.jgrapht)
   implementation(libs.jna)
   implementation(libs.jna.platform)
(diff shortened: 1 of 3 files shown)


kateliu20:
Looks like there is a [lint issue](https://googlesamples.github.io/android-custom-lint-rules/checks/GradleProjectIsolation.md.html):
`  /home/runner/work/foundry/foundry/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/topography/ModuleTopographyTask.kt:404: Error: Avoid using method getRootProject [GradleProjectIsolation from androidx.lint:lint-gradle]
            rootDirProperty.set(project.rootProject.rootDir)
`

jstewart5000 (approved):

PR #1301, opened 2025-03-31
Prevent Project Gen deadlock with invokeLater and a modeless dialog

kateliu20:
We've received several notes about project gen deadlocking newer versions of Android Studio, and here is the [issue](https://github.com/slackhq/foundry/issues/1288) for it.

I was able to reproduce the issue on main.

After the changes, I'm able to load the subproject without deadlock:

I wrote a few unit tests to confirm:
- `onOk` is called on `.doOKAction()` and `.dismissDialogAndSync()`.
- the Dialog is explicitly not modal.
- `onOk` is not called when cancelled.

### DIFF
--- a/platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/projectgen/ProjectGenMenuAction.kt
+++ b/platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/projectgen/ProjectGenMenuAction.kt
@@ -18,12 +18,14 @@ package foundry.intellij.skate.projectgen
 import com.intellij.openapi.actionSystem.ActionUpdateThread
 import com.intellij.openapi.actionSystem.AnAction
 import com.intellij.openapi.actionSystem.AnActionEvent
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.project.Project
 import foundry.intellij.skate.tracing.SkateSpanBuilder
 import foundry.intellij.skate.tracing.SkateTracingEvent
 import foundry.intellij.skate.util.getTraceReporter
 import foundry.intellij.skate.util.isProjectGenMenuActionEnabled
 import foundry.intellij.skate.util.isTracingEnabled
+import com.intellij.openapi.actionSystem.ActionManager
 import java.time.Instant
 
 class ProjectGenMenuAction : AnAction() {
@@ -31,7 +33,19 @@ class ProjectGenMenuAction : AnAction() {
     val currentProject = e.project ?: return
     if (!currentProject.isProjectGenMenuActionEnabled()) return
     val startTimestamp = Instant.now()
-    ProjectGenWindow(currentProject, e).show()
+
+    val dialog =
+      ProjectGenWindow(currentProject, e).apply {
+        onOk = {
+          ApplicationManager.getApplication().invokeLater {
+            val am = ActionManager.getInstance()
+            val sync = am.getAction("Android.SyncProject")
+            sync.actionPerformed(e)
+          }
+        }
+      }
+
+    dialog.show()
 
     if (currentProject.isTracingEnabled()) {
       sendUsageTrace(currentProject, startTimestamp)
--- a/platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/projectgen/ProjectGenMenuAction.kt
+++ b/platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/projectgen/ProjectGenMenuAction.kt
@@ -15,6 +15,7 @@
  */
 package foundry.intellij.skate.projectgen
 
+import com.intellij.openapi.actionSystem.ActionManager
 import com.intellij.openapi.actionSystem.ActionUpdateThread
 import com.intellij.openapi.actionSystem.AnAction
 import com.intellij.openapi.actionSystem.AnActionEvent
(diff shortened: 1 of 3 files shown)


platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/projectgen/ProjectGenMenuAction.kt
```diff
@@ -31,7 +33,19 @@ class ProjectGenMenuAction : AnAction() {
     val currentProject = e.project ?: return
     if (!currentProject.isProjectGenMenuActionEnabled()) return
     val startTimestamp = Instant.now()
-    ProjectGenWindow(currentProject, e).show()
+
+    val dialog =
```
kateliu20, line 37: `Android.SyncProject` was invoked immediately after showing a modal dialog with `ProjectGenWindow(currentProject, e).show()`. Modal dialogs block the EDT, and since `sync.actionPerformed(event)` also requires the EDT, the system could enter a deadlock.

I wrapped the sync logic with `invokeLater` to run the action after the dialog has been dismissed and the EDT is free again to prevent deadlock.

platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/projectgen/ProjectGenWindow.kt
```diff
@@ -27,17 +25,20 @@ import java.nio.file.Paths
...
     (currentProject.basePath?.let(Paths::get) ?: FileSystems.getDefault().getPath("."))
       .normalize()
       .also { check(Files.isDirectory(it)) { "Must pass a valid directory" } }
 
+  var onOk: (() -> Unit)? = null
+
   init {
     init()
     title = "Project Generator"
+    isModal = false
```
kateliu20, line 41: I set `isModal = false`, so the EDT stays unblocked and deferred work can proceed (refer to https://plugins.jetbrains.com/docs/intellij/threading-model.html#invoking-operations-on-edt-and-modality).

platforms/intellij/skate/src/test/kotlin/foundry/intellij/skate/projectgen/ProjectGenWindowTest.kt
```diff
@@ -0,0 +1,61 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package foundry.intellij.skate.projectgen
+
+import com.intellij.testFramework.fixtures.LightPlatformCodeInsightFixture4TestCase
+import org.junit.Test
+
+class ProjectGenWindowTest : LightPlatformCodeInsightFixture4TestCase() {
```
kateliu20, line 21: Running this whole test, I'm getting an error like this:

`Must pass a valid directory
java.lang.IllegalStateException: Must pass a valid directory`

But all these tests pass individually. Are there any suggestions here on how I can go about this?

platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/projectgen/ProjectGenMenuAction.kt
ZacSweers: Nice find!

platforms/intellij/skate/src/test/kotlin/foundry/intellij/skate/projectgen/ProjectGenWindowTest.kt
ZacSweers: Not sure off hand. Is there shared state we should be clearing between test runs?

ZacSweers (approved):

platforms/intellij/skate/src/test/kotlin/foundry/intellij/skate/projectgen/ProjectGenWindowTest.kt
kateliu20: After logging the project base path in each test, all tests share the same temporary project directory. This might be why it's failing when running the test class as a whole.

After creating a helper method `ensureProjectBasePathExists the class passed, so I call this in each test now. 

Does this approach make sense or is there something cleaner you recommend?

ZacSweers: Hmm, we should really just use a new temporary directory each run. Is there a reason it doesn't operate that way?

kateliu20: Based on [this](https://plugins.jetbrains.com/docs/intellij/light-and-heavy-tests.html) documentation, "Light tests reuse a project from the previous test run when possible." I think that's why it's reusing the directory. I used the light test fixture because I saw it's used in all the other tests, but should I change the test type in this case or leave as is?

platforms/intellij/skate/src/test/kotlin/foundry/intellij/skate/projectgen/ProjectGenWindowTest.kt
```diff
@@ -16,11 +16,24 @@
...
+    val basePath = project.basePath ?: return
+    val dir = File(basePath)
+    if (!dir.exists()) {
+      dir.mkdirs()
+    }
+  }
+
   @Test
   fun `test onOK callback is invoked on OK`() {
+    ensureProjectBasePathExists()
```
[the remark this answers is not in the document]
kateliu20: Sounds good, following the example from `SkateTracingReportTest` eb76e72cfbb2ab85b92ebfc2f4e93aeea165bb28, since I'm extending `LightPlatformCodeInsightFixture4TestCase` 🙇 thanks for the pointer

PR #1310, opened 2025-04-08
Remove guava from VersionNumber + add tests

ZacSweers:


kateliu20 (approved):
Nice demo in the highlights meeting too!

PR #1317, opened 2025-04-11
Add support for Metro DI in Foundry Gradle plugin

ZacSweers:
Introduced Metro as a dependency injection option for the Foundry Gradle plugin, alongside integration with Dagger and Anvil. This includes new configurations and properties to enable Metro and interop support for Dagger and Anvil.


.fleet/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-    "backend.maxHeapSizeMb": 1487
+    "backend.maxHeapSizeMb": 1487,
+    "gradle.autoImportOnSave": false
```
kateliu20, line 3: Can I ask what this is for and why you added it because of Metro?

kateliu20 (approved):

.fleet/settings.json
ZacSweers: Ah this is Fleet settings, was testing it out again while working on this

PR #1355, opened 2025-05-08
Use finalizeDsl if in android projects for DSL compat

ZacSweers:
This is necessary for KSP to work in AGP 8.10.x +

### DIFF
--- a/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/FoundryGradleUtil.kt
+++ b/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/FoundryGradleUtil.kt
@@ -26,6 +26,7 @@ import java.util.Locale
 import java.util.Optional
 import kotlin.io.path.ExperimentalPathApi
 import kotlin.io.path.deleteRecursively
+import org.gradle.api.Action
 import org.gradle.api.GradleException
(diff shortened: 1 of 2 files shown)


kateliu20 (approved):

platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/FoundryGradleUtil.kt
```diff
@@ -256,3 +257,17 @@ internal inline fun <reified T : Task> Project.namedLazy(
 
 internal fun <T : Any> Optional<T>.asProvider(providers: ProviderFactory) =
   providers.provider { get() }
+
+/**
+ * Executes the given action either upon the finalization of [AndroidComponentsExtension] DSL or
+ * after the project evaluation, providing compatibility for both scenarios.
+ *
+ * If we're in an Android project, we need to use the [AndroidComponentsExtension.finalizeDsl]
+ * callback to avoid https://github.com/google/ksp/issues/1789#issuecomment-2860100164
```
kateliu20, line 266: Just for my understanding, because of this error:
`It is too late to call addKspConfigurations as the callbacks already executed.`, we should use `finalizeDsl` so this callback can let us change these DSL objects?

PR #1358, opened 2025-05-13
Implement ProjectDependenciesDumpTask

ZacSweers:
This pull request introduces the `ProjectDependenciesDumpTask`, an alternative to `GenerateDependencyGraphTask`, where we'll generate simple direct deps and construct a dag ourselves outside of gradle

### DIFF
--- a/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/avoidance/ProjectDependenciesDumpTask.kt
+++ b/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/avoidance/ProjectDependenciesDumpTask.kt
@@ -1,0 +1,71 @@
+/*
+ * Copyright (C) 2025 Slack Technologies, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/avoidance/ProjectDependenciesDumpTask.kt
```diff
@@ -0,0 +1,71 @@
...
+      project.tasks.register<ProjectDependenciesDumpTask>(NAME) {
+        projectPaths.addAll(project.directProjectDependencies)
+        outputFile.set(project.layout.projectDirectory.file("gradle/implicit-projects.txt"))
+      }
+
+    private val Project.directProjectDependencies: Provider<Set<String>>
+      get() {
+        return provider {
+          configurations
+            .filter { c -> !c.isCanBeResolved }
```
autonomousapps, line 64: Note that you're filtering on unresolvable configurations. I had this in my example for a very specific use-case. You may want to get _all_ configurations or even only _resolvable_ configurations.

autonomousapps (approved):

PR #1367, opened 2025-05-16
Add support for all-projects.txt file in project gen

ZacSweers:
More support for Spotlight

### DIFF
--- a/platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/projectgen/models.kt
+++ b/platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/projectgen/models.kt
@@ -49,25 +49,37 @@ internal data class Project(
       feature.renderFiles(projectDir)
     }
 
-    val settingsFile = rootDir.resolve("settings-all.gradle.kts")
-    val includedProjects =
-      settingsFile
-        .readLines()
-        .asSequence()
-        .filter { it.startsWith("  \":") }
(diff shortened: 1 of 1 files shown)


platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/projectgen/models.kt
```diff
@@ -49,25 +49,37 @@ internal data class Project(
...
+          .plus(path)
+          .sorted()
+
+      settingsAllFile.writeText(
+        includedProjects.joinToString(
+          "\n",
+          prefix = "// Please keep these in alphabetical order!\ninclude(\n",
+          postfix = "\n)\n",
+        ) {
+          "  \"$it\","
```
kateliu20, line 70: Can you explain what this is for and what it would look like?

kateliu20 (approved):
🎉 so fast

platforms/intellij/compose/src/jvmMain/kotlin/foundry/intellij/compose/projectgen/models.kt
ZacSweers: Recommend reading the docs! https://github.com/joshfriend/spotlight

ZacSweers: But the block you commented on is what we do today, the older gradle style

PR #1406, opened 2025-06-17
Always defensively patch KSP tasks with sqldelight and viewbinding

ZacSweers:
In KSP 2.0.2 they've changed this a bit and now we need to just defensively patch this always


ZacSweers:
the CLA tool is busted, ignoring it

PR #1437, opened 2025-07-01
Encode values before going to disk

ZacSweers:
Just a little bit harder to read and also safer in xml. Added backward compatibility to it as well and started storing the version


kateliu20 (approved):
Is this more for safety because it's getting published to marketplace? Do we need to do this for Skate too?

PR #1440, opened 2025-07-02
Add project path dependency completions to Skate

ZacSweers:
This adds IDE support for project dependencies in Skate after moving off of gradle's type-safe project accessors. This sources known projects from `gradle/all-projects.txt`.

This was also a vibe coding exercise with claude code, so keep an eye out for odd AI issues.

**Highlights**
- Error lines if the project referenced isn't recognized
- Autocomplete for known projects
- Reference support (i.e. jump to) for paths to open their build file.

Filed [this issue](https://issuetracker.google.com/issues/429140552) separately to try to reduce noise from the Android plugin.

Some future work could include
- Making this work if sync fails (for some reason it doesn't currently, seemingly because the underlying PSI is deemed unindexed)
- Possibly supporting other sources of included projects (`settings.gradle`, etc)

image

image

image

### DIFF
--- a/platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/gradle/GradleProjectAnnotator.kt
+++ b/platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/gradle/GradleProjectAnnotator.kt
@@ -71,3 +71,23 @@ class GradleProjectAnnotator : Annotator {
     }
   }
 }
+
+/**
+ * Internal function to check if an element should be processed by the annotator. Exposed for
+ * testing.
+ */
+internal fun shouldAnnotatorProcessElement(
+  elementText: String,
+  hasProjectChildren: Boolean,
+): Boolean {
+  if (!elementText.contains("project(")) {
+    return false
+  }
+
+  // Skip if any child element also contains "project(" - let the child handle it
+  if (hasProjectChildren) {
+    return false
+  }
+
+  return true
+}
--- a/platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/gradle/GradleProjectAnnotator.kt
+++ b/platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/gradle/GradleProjectAnnotator.kt
@@ -20,6 +20,7 @@ import com.intellij.lang.annotation.Annotator
 import com.intellij.lang.annotation.HighlightSeverity
 import com.intellij.openapi.util.TextRange
 import com.intellij.psi.PsiElement
+import foundry.intellij.skate.SkatePluginSettings
 import java.util.regex.Pattern
 
 internal val PROJECT_CALL_PATTERN = Pattern.compile("""project\s*\(\s*["']([^"']+)["']\s*\)""")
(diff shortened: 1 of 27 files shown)


platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/gradle/GradleProjectAnnotator.kt
```diff
@@ -0,0 +1,93 @@
...
+    while (matcher.find()) {
+      val projectPath = matcher.group(1)
+      val startOffset = element.textRange.startOffset + matcher.start(1)
+      val endOffset = element.textRange.startOffset + matcher.end(1)
+      val range = TextRange.create(startOffset, endOffset)
+
+      if (!projectPathService.isValidProjectPath(projectPath)) {
+        // Create error annotation for invalid project path
+        holder
+          .newAnnotation(HighlightSeverity.ERROR, "Unknown project path: '$projectPath'")
```
kateliu20, line 66: This is nice!! Do you think it's possible to add the suggested project path or the format they should follow?

ZacSweers: not sure I follow, can you unpack that for me?

platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/gradle/GradleProjectCompletionContributor.kt
```diff
@@ -0,0 +1,107 @@
...
+import com.intellij.codeInsight.lookup.LookupElementBuilder
+import com.intellij.icons.AllIcons
+import com.intellij.patterns.PatternCondition
+import com.intellij.patterns.PlatformPatterns
+import com.intellij.patterns.PsiElementPattern
+import com.intellij.psi.PsiElement
+import com.intellij.util.ProcessingContext
+
+/** Provides autocomplete functionality for project(...) dependencies in Gradle build files. */
+class GradleProjectCompletionContributor : CompletionContributor() {
```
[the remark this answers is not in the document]
ZacSweers: None that I can see, any that you're specifically worried about?

platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/gradle/GradleProjectAnnotator.kt
kateliu20: I was thinking if a project path was written with typesafe project accessors format like:

`implementation(projects.tools.foundryCommon)`

if there should also be a suggestion / reminder to write it like this:
`implementation(project(":tools:foundry-common"))`

in case developers aren't used to this format or style? Or do you think it's unnecessary and is this mainly just for if the project path names don't match what's in the all projects file?

platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/gradle/GradleProjectCompletionContributor.kt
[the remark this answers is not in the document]
kateliu20: No, just might consider if it's reading from the `gradle/all-projects.txt` file if cached results aren't available

platforms/intellij/skate/src/main/kotlin/foundry/intellij/skate/gradle/GradleProjectAnnotator.kt
ZacSweers: ah no, this support isn't trying to help migrate people to project paths. We added a script to automatically do that for folks

PR #1471, opened 2025-07-23
Remove Copy Gradle Project Accessor from Skate

kateliu20:
Remove Skate "Copy Gradle Project Accessor" feature since we moved off of project accessors 

After:
Screenshot 2025-07-23 at 5 26 27 PM

Before:
Screenshot 2025-07-23 at 5 25 14 PM


ZacSweers:
I think it's fine to leave this actually, some small projects may use project accessors anyway

PR #1564, opened 2025-10-02
AGP 9 prep

ZacSweers:


kateliu20:
@ZacSweers would we have to migrate over to KSP2 based on this CLI warning?
`We noticed you are using KSP1 which is deprecated and support for it will be removed soon - please migrate to KSP2 as soon as possible. KSP1 will no longer be compatible with Android Gradle Plugin 9.0.0 (and above) and Kotlin 2.3.0 (and above)`

ZacSweers:
Yep

PR #1666, opened 2025-12-12
Update AGP to 8.13.2

jstewart5000:
### Description:
- Update AGP to `8.13.2`
- Adds support for Kotlin 2.3 
  - https://developer.android.com/build/releases/gradle-plugin#fixed-issues-agp-8.13.2


ZacSweers (approved):
lgtm, though worth mentioning you don't need to update it in foundry first to use it in slack-android since the latter imposes its own version

jstewart5000:
🤦🏾 Aw yeah good call, I was just thinking to get it in since we will have to release a new version anyways

PR #1848, opened 2026-06-08
fix: avoid "androidTestImplementation dependencies are ignored" warning

evanisnor:
## Problem

Library modules emit this warning at configuration time, once each, for nearly every Android library in a consuming project:

```
WARNING: androidTestImplementation dependencies are ignored because androidTest is disabled.
```

It fires even for modules that declare no `androidTest` dependencies of their own, which points at the convention plugin injecting them.

## Cause

The source is `applyPlatforms()`: it adds platform BOM dependencies to every configuration for which `isPlatformConfigurationName()` returns `true`. That matcher does **suffix** matching against the `RUNTIME` group, which contains `"implementation"`. Since `"androidTestImplementation"` ends with `"implementation"`, Foundry was platforming the `androidTestImplementation` configuration of every Android module.

Library modules do not enable the androidTest variant by default (it is opt-in via the `androidTest()` feature). When the variant is disabled, AGP ignores anything added to the `androidTest*` configurations and warns, hence the noise across the whole project.

## Fix

In `applyPlatforms()`, skip platforming `androidTest*` configurations (new `Configurations.isAndroidTest()` helper) unless the androidTest feature is enabled for the project. Modules that do opt into androidTest are unaffected and still get their platform dependencies.

## Verification

Tested against a large internal consumer via a composite build (`--rerun-tasks`, configuration cache disabled):

| Scenario | `androidTestImplementation ... ignored` warnings | Build |
|---|---|---|
| Library module, before fix | 23 | SUCCESSFUL |
| Same library module, after fix | 0 | SUCCESSFUL |
| androidTest-enabled module, after fix | 0 (no regression) | SUCCESSFUL |

### DIFF
--- a/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/Configurations.kt
+++ b/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/Configurations.kt
@@ -28,6 +28,13 @@ internal object Configurations {
 
   fun isTest(name: String): Boolean = testConfigurationRegex.matches(name)
 
+  /**
+   * Whether [name] is an android instrumentation test ("androidTest") configuration, such as
+   * `androidTestImplementation` or `androidTestUtil`. Note this intentionally does not match unit
+   * test configurations like `testImplementation`.
+   */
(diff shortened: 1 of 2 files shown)


ZacSweers (approved):
In the future can you de-fluff the AI descriptions? :)

PR #1910, opened 2026-07-16
Add `Project.androidTestProjectPaths()` accessor for androidTest projects

jstewart5000:
## Summary
Expose the set of androidTest-producing projects as a reusable public Foundry API, and publish the metadata that backs it unconditionally so any downstream consumer can enumerate those projects.

### Why
- Foundry knows which projects produce androidTest APKs, but there was no supported way for a build to obtain that set — consumers had to hand-maintain a list or run a task and read its output file.
- The backing `androidTestProjectMetadata` artifact was only published on the Fladle path.

### What Changed
- `StandardProjectConfigurations`: publish each affected project's `androidTestProjectMetadata` artifact unconditionally, instead of gating it behind `!enableEmulatorWtfForAndroidTest`. Only the Fladle APK-directory aggregation stays gated.
- `FoundryGradleUtil`: add `public fun Project.androidTestProjectPaths(): List<String>` — resolves the `SkippyAndroidTestProject` artifact view eagerly and reads each artifact's owning `ProjectComponentIdentifier` (project metadata, not file contents), so it is safe to call at configuration time and does not force the producer tasks to run. Fails fast with an actionable message if called before the root plugin's resolver exists.
- `Resolver`: extract `resolvableConfigurationName(declarableName)` so the accessor and the resolver share one configuration-name convention.

### Consumers
- Any build can now call `rootProject.androidTestProjectPaths()` to enumerate androidTest projects at configuration time.

### DIFF
--- a/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/FoundryGradleUtil.kt
+++ b/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/FoundryGradleUtil.kt
@@ -19,6 +19,8 @@ import com.android.build.api.variant.AndroidComponentsExtension
 import com.android.build.api.variant.ApplicationAndroidComponentsExtension
 import com.android.build.api.variant.LibraryAndroidComponentsExtension
 import com.android.build.api.variant.TestAndroidComponentsExtension
+import foundry.gradle.artifacts.FoundryArtifact
+import foundry.gradle.artifacts.Resolver
 import foundry.gradle.properties.gitExecProvider
 import foundry.gradle.properties.mapToBoolean
 import java.io.File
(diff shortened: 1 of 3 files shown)


platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/FoundryGradleUtil.kt
```diff
@@ -129,6 +132,33 @@ public fun Project.gitShaProvider(): Provider<String> {
...
+    configurations.findByName(configurationName)
+      ?: error(
+        "Configuration '$configurationName' not found. androidTestProjectPaths() must be called on " +
+          "the root project after Foundry's root plugin is applied, which registers the resolver " +
+          "this reads from."
+      )
+  // Only projects that publish the androidTest artifact match the view's attributes, so this is
+  // exactly the producer set.
+  return configuration.incoming
+    .artifactView { lenient(true) }
```
kateliu20, line 154: should this mirror the attribute post-filter the shared helper uses here? https://github.com/slackhq/foundry/blob/main/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/artifacts/Resolver.kt#L161 

Is `lenient(true)` enough to avoid picking up  non-matching artifacts?

PR #1923, opened 2026-07-27
Fix "androidTestImplementation dependencies are ignored"

slack-jallen:
All of our modules are showing this warning when compiled:

```
WARNING: androidTestImplementation dependencies are ignored because androidTest is disabled.
Add android.sync.suppressAgpWarnings=GENERIC to the gradle.properties file to suppress this warning.
```

### DIFF
--- a/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/Configurations.kt
+++ b/platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/Configurations.kt
@@ -37,6 +37,19 @@ internal object Configurations {
 
   fun isApi(name: String): Boolean = name.endsWith("api", ignoreCase = true)
 
+  /**
+   * Whether the androidTest configuration named [name] should be detached from its parent
+   * configurations (i.e. have its `extendsFrom` cleared).
+   *
+   * When a module hasn't opted into androidTest ([androidTestEnabled] is false), AGP disables the
+   * androidTest variant and builds nothing from its configurations. Those configurations still
+   * extend `implementation`/`api` though, so AGP reports every inherited dependency as ignored
+   * ("androidTestImplementation dependencies are ignored because androidTest is disabled").
+   * Detaching them leaves nothing to ignore. Opted-in modules keep the normal inheritance.
+   */
+  fun shouldDetachAndroidTestDependencies(name: String, androidTestEnabled: Boolean): Boolean =
+    isAndroidTest(name) && !androidTestEnabled
+
   object ErrorProne {
     const val ERROR_PRONE = "errorprone"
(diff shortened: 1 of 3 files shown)


ZacSweers:
Didn't @evanisnor already fix this in https://github.com/slackhq/foundry/pull/1848?

slack-jallen:
> Didn't @evanisnor already fix this in #1848?

Oh boy.. 1) we need to make a new foundry release. 2) @evanisnor I'm still seeing this error on latest foundry main (even with your changes), although much less.

```
> Configure project :features:legacy-csc
WARNING: androidTestImplementation dependencies are ignored because androidTest is disabled.
Add android.sync.suppressAgpWarnings=GENERIC to the gradle.properties file to suppress this warning.
w: ⚠️ Deprecated 'org.jetbrains.kotlin.android' plugin usage
The 'org.jetbrains.kotlin.android' plugin in project ':features:legacy-csc' is no longer required for Kotlin support since AGP 9.0.
Solution:
Remove both `android.builtInKotlin=true` and `android.newDsl=false` from `gradle.properties`, then migrate to built-in Kotlin.
See https://kotl.in/gradle/agp-built-in-kotlin for more details.
```

platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/Configurations.kt
```diff
@@ -37,6 +37,19 @@ internal object Configurations {
...
 
+  /**
+   * Whether the androidTest configuration named [name] should be detached from its parent
+   * configurations (i.e. have its `extendsFrom` cleared).
+   *
+   * When a module hasn't opted into androidTest ([androidTestEnabled] is false), AGP disables the
+   * androidTest variant and builds nothing from its configurations. Those configurations still
+   * extend `implementation`/`api` though, so AGP reports every inherited dependency as ignored
+   * ("androidTestImplementation dependencies are ignored because androidTest is disabled").
+   * Detaching them leaves nothing to ignore. Opted-in modules keep the normal inheritance.
```
ZacSweers, line 48: I don't think your LLM is correct here, otherwise every android project would have this issue all over. I think the problem is likely that we're still adding some deps automatically to those configs via the platform plugin and Evan's original PR maybe just missed a spot. Best thing to do on your end would be to inspect what dependencies are showing up on those configurations.

ZacSweers (commented):
Sidenote your LLM is leaking slack's internal PR template to this repo

platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/Configurations.kt
slack-jallen: @ZacSweers Thanks for taking a look! Hope you're doing well.

PR #1938, opened 2026-08-11
Enable Gradle 9.7 with project isolation

kaeawc:
## Summary
- Upgrade the wrapper to Gradle 9.7 and migrate root build configuration into project-local convention plugins.
- Enable isolated projects with strict configuration-cache regression coverage.
- Preserve explicit non-isolated paths for publishing and CLI ServiceLoader packaging.

## Validation
- `./gradlew clean check --quiet --configuration-cache-problems=fail --stacktrace`
- Repeat `./gradlew check --configuration-cache-problems=fail` to verify configuration-cache reuse
- `./gradlew verifyRoborazzi --quiet --configuration-cache-problems=fail --stacktrace`
- Isolated IntelliJ packaging and Dokka runs
- Non-isolated publication and CLI packaging checks


kaeawc:
I've been able to exercise this on Slack Android and with SqlDelight 2.4.0 snapshot gradle isolation passes

build.gradle.kts
```diff
@@ -84,57 +65,50 @@ if (file(".git").exists()) {
...
-      ":platforms:gradle.* -> :platforms:gradle.*",
-      ":platforms:intellij.* -> :platforms:intellij.*",
-      ":tools.* -> :tools.*",
+// Several plugins don't support isolated projects and need to be conditionally applied.
+val isolatedProjectsEnabled =
+  extensions.getByType<BuildFeaturesExtension>().isolatedProjects.getOrElse(false)
+
+// GraphAssert doesn't support isolated projects as it needs to traverse subproject configurations
+// See: https://github.com/jraska/modules-graph-assert/issues/XXX (TODO: file upstream issue)
+if (!isolatedProjectsEnabled) {
```
jstewart5000, line 74: The moduleGraphAssert plugin is applied only inside `if (!isolatedProjectsEnabled)`, but no CI job executes assertModuleGraph: isolated check never applies the plugin, and the non-isolated :tools:cli:jar/publish jobs never trigger the task Should we add a separate CI check like `./gradlew assertModuleGraph -Dorg.gradle.isolated-projects=false.`

tools/cli/build.gradle.kts
```diff
@@ -13,17 +13,29 @@
...
-  alias(libs.plugins.spotless)
+  id("foundry.maven-publish")
   alias(libs.plugins.moshix)
   alias(libs.plugins.kotlin.plugin.serialization)
-  alias(libs.plugins.ksp)
+  alias(libs.plugins.ksp) apply false
+}
+
+if (!isolatedProjectsEnabled) {
+  apply(plugin = "com.google.devtools.ksp")
```
jstewart5000, line 38: Since we are skipping KSP plugin here, a developer running the documented `publishToMavenLocal` without the flag ships a broken artifact. Should we add a configuration-time `logger.warn` or task failure when isolation is enabled so the missing descriptor is diagnosable.

tools/cli/build.gradle.kts
```diff
@@ -70,5 +84,7 @@ dependencies {
   testImplementation(libs.kaml)
   testImplementation(libs.truth)
 
-  ksp(libs.autoService.ksp)
+  if (!isolatedProjectsEnabled) {
+    "ksp"(libs.autoService.ksp)
```
jstewart5000, line 88: Same as above: Since we are skipping KSP plugin here, a developer running the documented `publishToMavenLocal` without the flag ships a broken artifact. Should we add a configuration-time `logger.warn` or task failure when isolation is enabled so the missing descriptor is diagnosable.

platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/FoundryGradleUtil.kt
```diff
@@ -239,18 +236,9 @@ internal inline fun <reified T : Task> Project.namedLazy(
     return
   } catch (ignored: UnknownTaskException) {}
 
-  var didRun = false
-
-  tasks.withType(T::class.java) {
+  tasks.withType(T::class.java).configureEach {
```
jstewart5000, line 239: Claude: The `afterEvaluate/didRun` guard that made `namedLazy` throw when a target task was never registered was removed below, so a renamed/removed/mistyped task name now silently skips the action. The change is understandable for isolation compatibility, but should we restore a config-cache-safe failure path?

build.gradle.kts
```diff
@@ -161,266 +135,6 @@ if (!spotlightEnabled) {
...
-    val configToAdd =
-      if (pluginManager.hasPlugin("org.jetbrains.kotlin.multiplatform")) {
-        "jvmMainCompileOnly"
-      } else {
-        "compileOnly"
-      }
-    afterEvaluate { project.dependencies.add(configToAdd, libs.slackLints.annotations) }
-  }
-}
-
```
jstewart5000, line 423: Do we need to add a `foundry.lint` convention plugin

.github/workflows/ci.yml
```diff
@@ -76,8 +76,8 @@ jobs:
           with:
             lfs: 'true'
 
-        - name: Check LFS files
-          uses: actionsdesk/lfs-warning@v3.3
+        - name: Verify LFS files
+          run: git lfs fsck
```
jstewart5000, line 80: Any specific reason we moved away from `actionsdesk/lfs-warning@v3.3` (flags large files committed outside LFS)?

build-logic/src/main/kotlin/foundry.kotlin-jvm.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+      "-Xjdk-release=${jvmTargetVersion.target}",
+    )
+    optIn.addAll(
+      "kotlin.contracts.ExperimentalContracts",
+      "kotlin.experimental.ExperimentalTypeInference",
+      "kotlin.ExperimentalStdlibApi",
+      "kotlin.time.ExperimentalTime",
+    )
+  }
+}
```
jstewart5000, line 66: I think outside of:
```
languageVersion.set(KotlinVersion.DEFAULT)
apiVersion.set(KotlinVersion.DEFAULT)
```

This block of code is close to identical in:
- foundry.kotlin-jvm.gradle.kts:44
- foundry.kotlin-jvm-gradle.gradle.kts:50
- foundry.kotlin-jvm-intellij.gradle.kts:43

Could we extract `internal fun Project.configureFoundryKotlinJvm(...)` into a shared `.kt` in build-logic; each script passes its three differing params for simplification.

build-logic/src/main/kotlin/foundry.kotlin-jvm-gradle.gradle.kts
```diff
@@ -0,0 +1,81 @@
...
+plugins {
+  id("org.jetbrains.kotlin.jvm")
+  id("org.jetbrains.kotlin.plugin.sam.with.receiver")
+  id("com.squareup.sort-dependencies")
+}
+
+val catalog = extensions.getByType<VersionCatalogsExtension>().named("libs")
+val jvmTargetVersion =
+  catalog.findVersion("jvmTarget").get().toString().let { JvmTarget.fromTarget(it) }
+val jdkVersion = catalog.findVersion("jdk").get().toString().toInt()
```
jstewart5000, line 33: Do we need to restore `.removeSuffix("-ea")` before `.toInt()`
```
catalog.findVersion("jdk").get().toString().removeSuffix("-ea").toInt()
```

build-logic/src/main/kotlin/foundry.kotlin-jvm-intellij.gradle.kts
```diff
@@ -0,0 +1,72 @@
...
+
+plugins {
+  id("org.jetbrains.kotlin.jvm")
+  id("com.squareup.sort-dependencies")
+}
+
+val catalog = extensions.getByType<VersionCatalogsExtension>().named("libs")
+val jvmTargetVersion =
+  catalog.findVersion("jvmTargetIdea").get().toString().let { JvmTarget.fromTarget(it) }
+val jdkVersion = catalog.findVersion("jdk").get().toString().toInt()
```
jstewart5000, line 30: Do we need to restore `.removeSuffix("-ea")` before `.toInt()`
```
catalog.findVersion("jdk").get().toString().removeSuffix("-ea").toInt()
```

build-logic/src/main/kotlin/foundry.kotlin-jvm.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+
+plugins {
+  id("org.jetbrains.kotlin.jvm")
+  id("com.squareup.sort-dependencies")
+}
+
+val catalog = extensions.getByType<VersionCatalogsExtension>().named("libs")
+val jvmTargetVersion =
+  catalog.findVersion("jvmTarget").get().toString().let { JvmTarget.fromTarget(it) }
+val jdkVersion = catalog.findVersion("jdk").get().toString().toInt()
```
jstewart5000, line 31: Do we need to restore `.removeSuffix("-ea")` before `.toInt()`
```
catalog.findVersion("jdk").get().toString().removeSuffix("-ea").toInt()
```

build.gradle.kts
```diff
@@ -84,57 +65,50 @@ if (file(".git").exists()) {
...
-      ":platforms.* -> :tools.*",
-      ":platforms:gradle.* -> :platforms:gradle.*",
-      ":platforms:intellij.* -> :platforms:intellij.*",
-      ":tools.* -> :tools.*",
+// Several plugins don't support isolated projects and need to be conditionally applied.
+val isolatedProjectsEnabled =
+  extensions.getByType<BuildFeaturesExtension>().isolatedProjects.getOrElse(false)
+
+// GraphAssert doesn't support isolated projects as it needs to traverse subproject configurations
+// See: https://github.com/jraska/modules-graph-assert/issues/XXX (TODO: file upstream issue)
```
jstewart5000, line 73: Is this upstream issue already filed for this one? Or just temporary placeholder?

platforms/gradle/foundry-gradle-plugin/build.gradle.kts
```diff
@@ -50,16 +53,26 @@ buildConfig {
...
-  from(rootProject.layout.projectDirectory.dir("config/git/hooks")) {
+  from(rootDir.resolve("config/git/hooks")) {
     // Give it a common prefix for us to look for
     rename { name -> "githook-$name" }
   }
 }
 
 moshi { enableSealed.set(true) }
 
+lint { baseline = file("lint-baseline.xml") }
```
jstewart5000, line 64: I think it was deleted on purpose, but points at a baseline file this PR deletes

platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/FoundryBasePlugin.kt
```diff
@@ -115,7 +115,11 @@ internal class FoundryBasePlugin @Inject constructor(private val buildFeatures:
     ProjectDependenciesDumpTask.register(target)
     target.configureClasspath(foundryProperties)
 
-    if (!this.buildFeatures.isolatedProjects.requested.getOrElse(false)) {
+    if (!target.isRootProject) {
+      globalConfig.configureProject(target)
+    }
+
+    if (!this.buildFeatures.isolatedProjects.active.getOrElse(false)) {
```
jstewart5000, line 122: Claude: With root `subprojects {}/allprojects {}` removed and `configureSpotless` skipped at root under isolation (`FoundryBasePlugin.kt:122`), conventions reach a module only through its explicit `id("foundry.spotless")`/base-plugin lines. A new module that forgets them silently gets zero enforcement, and `configureProject` (`GlobalConfig.kt:45-46`) emits no diagnostic when a configurer matches no projects. Consider a root verification task (isolation-disabled) asserting every source module opts into the expected convention plugins.

jstewart5000 (changes requested):
Overall looked good, just a couple things for additional logging / safe guards and some dead code. Some I just had overall questions for and not necessarily asking for a change.

build.gradle.kts
kaeawc: Yeah we should add one, should be straightfoward enough.

tools/cli/build.gradle.kts
kaeawc: Yeah I'll make something cc safe

tools/cli/build.gradle.kts
kaeawc: Looks like KSP latest is IP compliant now - exploring that

build.gradle.kts
kaeawc: Working on addressing this

build-logic/src/main/kotlin/foundry.kotlin-jvm-intellij.gradle.kts
kaeawc: Yup restoring this, thanks good catch

platforms/gradle/foundry-gradle-plugin/src/main/kotlin/foundry/gradle/FoundryBasePlugin.kt
kaeawc: Working on this

build.gradle.kts
kaeawc: added actual issue. we could move to a fork that actually supports IP, maybe in a follow up

